PR #4, opened 2018-05-07
Also read reweighting info etc.

iamholger:
LHEInit and LHEProcIfno are some additional things of the lhef3 standard (according to Stefan Prestel).
They re so simple, I just left them as dicts.

The LHEEvent now also has attributes and weights. The attributes are supposedly abused for reweighting information.

In order to not mess up things to badly, I made a copy of your readLHE and extended it to also fill the weights and attributes properties.

Finally, readNumEvents returns the number of events in the file.

### DIFF
--- a/pylhe/__init__.py
+++ b/pylhe/__init__.py
@@ -1,13 +1,15 @@
 import os
 
 class LHEFile(object):
-    def __init__(self):
-        pass
+  def __init__(self):
+    pass
 
 class LHEEvent(object):
-    def __init__(self,eventinfo,particles):
-        self.eventinfo = eventinfo
-        self.particles = particles
+    def __init__(self,eventinfo,particles,weights=None,attributes=None):
+        self.eventinfo  = eventinfo
+        self.particles  = particles
+        self.weights    = weights
+        self.attributes = attributes
         for p in self.particles:
             p.event = self
 
@@ -18,11 +20,11 @@ class LHEEventInfo(object):
             raise RuntimeError
         for k,v in kwargs.iteritems():
             setattr(self,k,v)
-    
+
     @classmethod
     def fromstring(cls,string):
         return cls(**dict(zip(cls.fieldnames,map(float,string.split()))))
-    
+
 
 class LHEParticle(object):
     fieldnames = fieldnames = ['id','status','mother1','mother2','color1','color2','px','py','pz','e','m','lifetime','spin']
@@ -31,12 +33,12 @@ class LHEParticle(object):
             raise RuntimeError
         for k,v in kwargs.iteritems():
             setattr(self,k,v)
-    
+
     @classmethod
     def fromstring(cls,string):
         obj = cls(**dict(zip(cls.fieldnames,map(float,string.split()))))
         return obj
-    
+
     def mothers(self):
         mothers = []
         first_idx  =  int(self.mother1)-1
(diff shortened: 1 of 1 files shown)


pylhe/__init__.py
```diff
@@ -1,13 +1,15 @@
 import os
```
lukasheinrich: add logging imports and set up

```python
import logging
log = logging.getLogger(__name__)
```

pylhe/__init__.py
```diff
@@ -31,27 +33,94 @@ def __init__(self, **kwargs):
...
+        if element.tag == "initrwgt":
+            initDict["weightgroup"]={}
+            for child in element:
+                # Find all weightgroups
+                if child.tag=="weightgroup" and child.attrib!={}:
+                    try:
+                        wg_type = child.attrib["type"]
+                    except Exception, e:
+                        print "weightgroup must have attribute 'type'"
+                        sys.exit(1)
```
lukasheinrich: Is there a reason we need to exit hard here?  The raised exception should already make the progam halt.  Also the exception should probably we narrowed down to, I assume, `KeyError`. If we want to add a meaningful message we can do

```
                 try:
                  except KeyError:
                     log.exception("weightgroup must have attribute 'type'")
```

pylhe/__init__.py
```diff
@@ -31,27 +33,94 @@ def __init__(self, **kwargs):
...
+                    _temp = {"attrib":child.attrib}
+                    _temp["weights"] = {}
+                    # Iterate over all weights in this weightgroup
+                    for w in child:
+                        if not w.tag=="weight": continue
+                        try:
+                            wg_id = w.attrib["id"]
+                        except Exception, e:
+                            print "weight must have attribute 'id'"
+                            sys.exit(1)
```
lukasheinrich: same as comment above

pylhe/__init__.py
```diff
@@ -60,11 +129,50 @@ def readLHE(thefile):
...
+            if sub.tag=="rwgt":
+                for r in sub:
+                    if r.tag=="wgt":
+                        eventdict["weights"][r.attrib["id"]] = float(r.text.strip())
+        # yield eventdict
+        yield LHEEvent(eventdict["eventinfo"], eventdict["particles"], eventdict["weights"], eventdict["attrib"])
+
+  except ET.ParseError:
+    print "WARNING. Parse Error."
+    return
```
lukasheinrich: we should change this to `log.warning(..)` instead of print

lukasheinrich:
Thanks @iamholger for the PR. I added some small comments, mainly abbout logging and exception handling otherwise this looks nice!

pylhe/__init__.py
iamholger: no, should have ben an exception in the first place, sorry

pylhe/__init__.py
iamholger: cool!

iamholger:
Tahnks so much!

lukasheinrich:
why did you close the PR? :)

marinang:
These new features are really helpful actually to recover the weights from the lhefile. 
I was using Pythia python bindings to shower events from a lhefile but you cannot recover the weights with the current bindings ....

lukasheinrich:
@iamholger do we want to revive this PR? @marinang if you already have a working version maybe you can push it upstream?

marinang:
@lukasheinrich for sure I cannot push in this PR, let's wait a bit  if @iamholger  wants to revive it then I can give some suggestions.

iamholger:
I think this is a good idea --- in principle we could
aggregate the hdf5 based scripts here, too.

Holger

lukasheinrich:
@marinang have you added any patches to the original PR? maybe we can integrate them or merge @iamholger's PR first and add your patches on that new `master`

pylhe/__init__.py
```diff
@@ -31,27 +33,94 @@ def __init__(self, **kwargs):
...
+            data = element.text.split('\n')[1:-1]
+            initDict["initInfo"]=LHEInit.fromstring(data[0])
+            initDict["procInfo"]=[LHEProcInfo.fromstring(d) for d in data[1:]]
+        if element.tag == "initrwgt":
+            initDict["weightgroup"]={}
+            for child in element:
+                # Find all weightgroups
+                if child.tag=="weightgroup" and child.attrib!={}:
+                    try:
+                        wg_type = child.attrib["type"]
```
marinang: I have only changed here the tag, in my lhefiles produces by MadGraph it's `name` and not `weightgroup`.

marinang:
@lukasheinrich So I think I could be merged as it is now and I can just implement small patches in another PR.

lukasheinrich:
hi @marinang I merged this... feel free to add your patch.  thanks @iamholger for the PR :)

PR #8, opened 2019-07-23
Update __init__.py

ariaradick:
Line 19 should read "for k,v in kwargs.items():" rather than "for k,v in kwargs.iteritems():" in order to be compatible with Python 3. See the answer to [this stack exchange question.](https://stackoverflow.com/questions/8899129/how-do-i-loop-through-kwargs-in-python)


lukasheinrich:
awesome, thanks!

PR #15, opened 2019-11-08
Add CI with GitHub Actions

matthewfeickert:
Add skeleton of CI with GitHub Actions to be filled in with pytest tests later.

Also fix some Python 3 issues and add missing dependencies.

### DIFF
--- a/pylhe/__init__.py
+++ b/pylhe/__init__.py
@@ -117,7 +117,7 @@ def loads():
 def readLHEInit(thefile):
     """
     Read the init blocks. Return dict. This encodes the weight group
-    and related things according to https://arxiv.org/pdf/1405.1067.pdf
+    and related things according to https://arxiv.org/abs/1405.1067
     This function returns a dict.
     """
     initDict = {}
--- a/pylhe/__init__.py
+++ b/pylhe/__init__.py
@@ -134,7 +134,7 @@ def readLHEInit(thefile):
                     try:
                         wg_type = child.attrib["type"]
                     except Exception, e:
-                        print "weightgroup must have attribute 'type'"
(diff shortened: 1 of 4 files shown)


matthewfeickert:
@lukasheinrich This is ready for review.

pylhe/__init__.py
```diff
@@ -133,8 +134,8 @@ def readLHEInit(thefile):
                 if child.tag == "weightgroup" and child.attrib != {}:
                     try:
                         wg_type = child.attrib["type"]
-                    except Exception, e:
-                        print "weightgroup must have attribute 'type'"
+                    except Exception:
+                        print("weightgroup must have attribute 'type'")
                         sys.exit(1)
```
lukasheinrich: can we just re-raise here instead of exiting?

pylhe/__init__.py
```diff
@@ -144,8 +145,8 @@ def readLHEInit(thefile):
                             continue
                         try:
                             wg_id = w.attrib["id"]
-                        except Exception, e:
-                            print "weight must have attribute 'id'"
+                        except Exception:
+                            print("weight must have attribute 'id'")
                             sys.exit(1)
```
lukasheinrich: dito

pylhe/__init__.py
lukasheinrich: probably should be KeyError

matthewfeickert: Yes, definitely. These are just the things that were in before from the other PR (with a quick edit to get pyflakes passing).

matthewfeickert: Updated.

lukasheinrich:
awesome.. thanks @matthewfeickert!

PR #16, opened 2019-11-08
Use src directory structure for project

matthewfeickert:
Requires #15

Move to using a `src` directory structure for the project and then add a `MANIFEST.in`


matthewfeickert:
@lukasheinrich this is ready for review.

lukasheinrich:
you're on a roll :rocket:

matthewfeickert:
> you're on a roll

:) Though I think I'll stop here for a while, as the next thing to do is add unit tests and identify files that make sense to use on them. We can follow up next week &mdash; have a good flight home!

PR #28, opened 2020-05-16
Add option to read optional event information

8me:
An option to read the optional event information with `readLHEWithAttributes` and a corresponding field in `LHEEvent` was added. This optional event information is defined in the Les Houches file format according to [A Standard format for Les Houches event files](https://inspirehep.net/literature/725284) section 4.

### DIFF
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -12,11 +12,17 @@ class LHEFile(object):
 
 
 class LHEEvent(object):
-    def __init__(self, eventinfo, particles, weights=None, attributes=None):
+    def __init__(self,
+                 eventinfo,
+                 particles,
+                 weights=None,
+                 attributes=None,
+                 optional=None):
         self.eventinfo = eventinfo
         self.particles = particles
         self.weights = weights
         self.attributes = attributes
+        self.optional = optional
         for p in self.particles:
             p.event = self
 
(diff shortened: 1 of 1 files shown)


src/pylhe/__init__.py
```diff
@@ -191,20 +199,22 @@ def readLHEWithAttributes(thefile):
                 for p in particles:
                     if not p.strip().startswith("#"):
                         eventdict["particles"] += [LHEParticle.fromstring(p)]
+                    elif parse_optional:
+                        if "optional" not in list(eventdict.keys()):
+                            eventdict["optional"] = []
```
lukasheinrich, line 204: maybe this could be added as a default empty list in the eventdict, regardless of the `parse_optional`?

lukasheinrich:
Thanks @8me! Can you explain a bit in which usecases this type of optional information appears?

8me:
I discovered this usecase when parsing data from the [GiBUU](https://gibuu.hepforge.org/) neutrino generator. The test file [EventOutput.Pert.00000001.lhe](https://github.com/8me/pylhe/blob/testing/tests/test_data/EventOutput.Pert.00000001.lhe) includes an extra optional line per event for the related neutrino information so to say.
As this is also included in the format description of LHE I thought this should be more a feature of `pylhe` than some individual one in my analysis code.

src/pylhe/__init__.py
8me: See [1c8e8df](https://github.com/scikit-hep/pylhe/pull/28/commits/1c8e8df50922317df9fc88369c912df5d8fade28)

8me:
In order to keep this PR alive ... is there anything left to do from your point of view, @lukasheinrich?

lukasheinrich:
the only thing that was left to do was to merge it :) Thanks again for the nice PR!

PR #27, opened 2020-05-13
feat: Use particle for labeling in visualizations

fuenfundachtzig:
* pypdt's particle() will return None for SUSY PDG IDs in 0.7.5
* backslash needs to be escaped twice
* networkx 1.11 has write_dot in submodule (no guarantees about this fix)

PR #44, opened 2020-09-30
ci: Publish releases to PyPI through GitHub releases

matthewfeickert:
This PR uses publication type releases to publish to PyPI by running the publication workflow on `published` `release` events and then verifying that the PyPI upload step is running on one. This follows what I'll call the `corner` model (which I now think is how GitHub Actions assumed people would do things): 

This follows the example that @dfm has setup in [`corner`](https://github.com/dfm/corner.py) with running the workflow on [`published` `release` events](https://github.com/dfm/corner.py/blob/efd25380ee121c4d565ed9aff2f98e184cc9dcef/.github/workflows/tests.yml#L7-L8) and then only running the [publish to PyPI step on that conditions](https://github.com/dfm/corner.py/blob/efd25380ee121c4d565ed9aff2f98e184cc9dcef/.github/workflows/tests.yml#L82). To instigate this workflow one would do the following:

- Locally on `master` run:

```
$ git checkout master && git pull # verify that you're on master and synced with GitHub
$ bump2version <part> # bump the version and create a commit and tag
$ git push origin master --tags # push the commit and the tag to GitHub, causing TestPyPI to publish
```
- Then on GitHub:
   1. Go to releases: https://github.com/scikit-hep/pylhe/releases
   2. Click "Draft a new release"
   3. On the new page enter the tag you just pushed (e.g. `v0.1.0`) in the "Tag version" box and the "Release title" box (to make it easy unless you really want to get descriptive)
   4. Enter any release notes and click "Publish release"
- This then kicks of the publication CD workflow that will use the PyPI API key to publish.

This is a bit different from the `pyhf` way of doing things, but also lets things run through on TestPyPI on the tag before things are ever released.

Let me know your thoughts @lukasheinrich.

```
* Publish to PyPI through published release events triggered through GitHub releases
   - c.f. https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -8,6 +8,8 @@ on:
   pull_request:
     branches:
     - master
+  release:
+    types: [published]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -66,7 +66,7 @@ jobs:
         password: ${{ secrets.test_pypi_password }}
         repository_url: https://test.pypi.org/legacy/
     - name: Publish distribution 📦 to PyPI
-      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && github.repository == 'scikit-hep/pylhe'
+      if: github.event_name == 'release' && github.event.action == 'published' && github.repository == 'scikit-hep/pylhe'
       uses: pypa/gh-action-pypi-publish@v1.3.1
       with:
         password: ${{ secrets.pypi_password }}


matthewfeickert:
@lukasheinrich Once this PR is resolved or changed we should probably also release `v0.0.6` to get in the recent contribution.

eduardo-rodrigues:
My 2 cents but the enhancements and changes of dependencies warrant a 0.1.0 I reckon :-).

matthewfeickert:
> My 2 cents but the enhancements and changes of dependencies warrant a 0.1.0 I reckon :-).

I would normally agree, but `v0.0.4` and `v0.0.5` were both breaking changes, so at this point it seems like keeping things at `v0.0.X` until there is some more stability in the library structure is maybe a bit more clear in terms of broadcasting "this is all very unstable" given the double `0`.

matthewfeickert:
@eduardo-rodrigues I'm going to end up agreeing with you here on the `v0.1.0` given that PR #27 changed install requirements.

eduardo-rodrigues:
As you wish @matthewfeickert. I see things here the opposite way :D - the change of dependencies is something to flag to a user, hence the explicit minor update to 0.1.0, whereas the technical/CI updates of the issues you mention are something transparent and/or really minor as far as a user is concerned, hence seems like a patch or bug fix release of the 3rd digit. This is al largely academic in the end, so ... I look forward to seeing a version 0.1.0, though, which will flag a library more ready for us, and that's the important thing :).

Issue #43, opened 2020-09-28
Drop Python 2.7 support


matthewfeickert:
Following the discussion RE: how to properly drop support for a release in https://github.com/scikit-hep/pyhf/issues/1075, it seems that as only Python 3.6+ is tested in CI now that either Python 2.7 support should be dropped or we should test Python 2.7 make a statement about how long Python 2.7 support will be provided in the `README`. Either way we should add a `python_requires` to `setup.cfg`.

Thoughts @lukasheinrich?

matthewfeickert:
What I would actually recommend doing is cutting a release `v0.1.0` that has `python_requires >= 2.7` and then as soon as that is out cut release `v0.2.0` that has `python _requires >= 3.6`.

lukasheinrich:
@matthewfeickert that sound pretty reasonable to me!

matthewfeickert:
> that sound pretty reasonable to me!

Okay. Once the CI unfreezes for MacOS tests I'll setup PRs and then also make releases.

eduardo-rodrigues:
Yep. See also our org guidelines/thoughts that we had put up some months ago - https://scikit-hep.org/supported-python-versions.

eduardo-rodrigues:
This being said, it would be a *big* shame if the nice improvements/enhancements in https://github.com/scikit-hep/pylhe/pull/27 do not make it to the 1.0.0 release. I don't think there's much work on this PR to be done, to get it merged ... I can help for the parts related to `Particle`, BTW.

matthewfeickert:
> Yep. See also our org guidelines/thoughts that we had put up some months ago

Yup I knew about them and I agree sticking to those is important. As @lukasheinrich better who some of the theory people are who are currently using `pylhe` than I do I wanted to make sure that we didn't break things for them, but making a `v0.1.0` release with `python_requires>=2.7` will ensure that there is a release that we can have for them if they're stuck for whatever reason.

> not make it to the 1.0.0 release

I think that a `v1.0.0` release is a _long_ way off, as we're all hesitant to freeze the API (and the library structure itself needs work as well). 

> I don't think there's much work on this PR to be done, to get it merged

We should probably discuss with the PR authors on the PR if they are going to come back to it, or if we should rebase it ourselves and then finish it off to get it in.

hugovk:
[...]

PR #47, opened 2020-11-02
fix: Require Python 2.7 or later or Python 3.6+

matthewfeickert:
Partially addresses Issue #43 

In preparation of `v0.1.0` drop Python 2.7 support and support for Python 3 older than 3.6

```
* Require Python 2.7 or later or Python 3.6+ for install
* Add Python 2.7 to PyPI metadata
   - To be dropped after v0.1.0 is published
* Temporarily loosen networkx requirements to >=2.2 to allow for Python 2.7 install
   - Test Python 2.7 until v0.1.0 is published
* Temporarily explicitly wrap paths as str and remove use of f-strings until Python 2 dropped
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ package_dir =
     = src
 packages = find:
 include_package_data = True
-python_requires = >=2.7
+python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*
 install_requires =
     networkx>=2.4
     tex2pix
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -35,8 +35,9 @@ package_dir =
     = src
 packages = find:
 include_package_data = True
+python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*
 install_requires =
-    networkx>=2.4
+    networkx>=2.2
```
matthewfeickert, line 40: Hi @fuenfundachtzig :wave:. In PR #27 you set the minimum required version to be `2.4`, but the last supported version of `networkx` under Python 2.7 was `2.2`. We don't want to break Python 2.7 until we can release `v0.1.0` and then drop Python 2 support (Issue #43), so do you know of something in particular that _needs_ `networkx` `v2.4`? Or was this more a recommendation?

matthewfeickert: cc @cburton12 as well if they know.

PR #64, opened 2020-12-17
build: Use Particle v0.14 for minimal compatible release

eduardo-rodrigues:
Trivial update to latest series. Not that the updates will matter to you here, but no harm in using the latest and greatest.
(In fact explicit support for Python 3.9 only came with the 0.13 series, so better update in any case.)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ python_requires = >=3.6
 install_requires =
     networkx~=2.2
     tex2pix~=0.3
-    particle~=0.12
+    particle~=0.14
 
 [options.packages.find]
 where = src


matthewfeickert:
> Trivial update to latest series. Not that the updates will matter to you here, but no harm in using the latest and greatest.
> (In fact explicit support for Python 3.9 only came with the 0.13 series, so better update in any case.)

Thanks for the PR @eduardo-rodrigues, but with [the compatible release syntax ](https://github.com/scikit-hep/pylhe/pull/62) `v0.14` is already being picked up. Example:

```
$ docker run --rm -it python:3.8 /bin/bash
root@b53098b9144a:/# python -m pip install "particle~=0.12"
root@b53098b9144a:/# python -m pip list | grep particle
particle   0.14.0
```

So setting

https://github.com/scikit-hep/pylhe/blob/51153ca297c0a89d76b7184dc56c980955555ff9/setup.cfg#L43

is just requiring that `pip` have a lower bound of `v0.12` and to be within the `v0.X` release series.

I agree there's no harm in merging this though.

matthewfeickert (approved):
:shipit:

eduardo-rodrigues:
I see, thanks for pointing that out.

PR #69, opened 2021-02-06
build: Remove universal wheel flag

matthewfeickert:
```
* Remove universal wheel setting as pylhe is Python 3 only
   - Fixes py2.py3 tag naming scheme for wheels
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,6 @@ classifiers =
     Programming Language :: Python :: 3.9
 
 [bdist_wheel]
-universal = 1
 
 [options]
 setup_requires =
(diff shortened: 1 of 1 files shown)


matthewfeickert:
For posterity (and self reference), in the [PyPA's "Packaging and distributing projects" guide the section on wheels](https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels) notes that

> Only use the `--universal setting`, if:
>
> 1. Your project runs on Python 2 and 3 with no changes (i.e. it does not require 2to3).
> 2. Your project does not have any C extensions.

PR #74, opened 2021-03-31
feat: Add support for reading compressed files

### LINKED ISSUE #65: Include ability to read gzip'ed input LHE file
As a quick suggestion for an improvement, it might be very handy to include the ability of opening zipped lhe files without the need for unzipping beforehand.
This can be achieved checking the file type and in case using something like
https://stackoverflow.com/questions/10566558/python-read-lines-from-compressed-text-files

This feature can be then disabled if the right dependency is not there, but it looks to me a rather minimal dependency and a quite wide usage for this feature.

 Thanks for considering it.

matthewfeickert:
Resolves #65 

```
* Add support for reading gzip-ed files by first opening gzip files if passed in to the readLHE APIs
   - Use _extract_fileobj with a context manager for the readLHE APIs
* Add tests for reading behavior 

Co-authored-by: Lukas <lukas.heinrich@gmail.com>
```

### DIFF
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -4,6 +4,7 @@ import networkx as nx
 import tex2pix
 import subprocess
 from particle.converters.bimap import DirectionalMaps
+import gzip
 
 
 class LHEFile:
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -4,7 +4,6 @@ import networkx as nx
 import tex2pix
 import subprocess
 from particle.converters.bimap import DirectionalMaps
-import gzip
 
 
 class LHEFile:
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -130,11 +130,15 @@ def _open_gzip_file(filepath):
     if not Path(filepath).name.lower().endswith(".gz"):
         return filepath
 
-    with gzip.open(filepath, "r") as gzip_file:
-        try:
-            gzip_file.read(1)
-        except (OSError, gzip.BadGzipFile) as err:
-            raise err(f"Input file {filepath} is not a compressed file.\n")
+    with open(filepath, "rb") as gzip_file:
+        header = gzip_file.read(2)
+
+    gzip_magic_number = b"\x1f\x8b"
+    if header != gzip_magic_number:
+        raise OSError(
+            f"Input file {filepath} is not a compressed file.\n"
+            + f"{filepath} has header of {header} and not gzip's {gzip_magic_number}.\n"
+        )
     return gzip.open(filepath, "r")
 
 
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -2,6 +2,7 @@ import gzip
 import os
 import subprocess
 import xml.etree.ElementTree as ET
+from pathlib import Path
 
 import networkx as nx
 import tex2pix
@@ -126,7 +127,7 @@ def _open_gzip_file(filepath):
     """
     Checks to see if a file is compressed, and if so, open it with gzip
     """
-    if not filepath.lower().endswith(".gz"):
+    if not Path(filepath).name.lower().endswith(".gz"):
         return filepath
 
     with gzip.open(filepath, "r") as gzip_file:
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -132,13 +132,13 @@ def _open_gzip_file(filepath):
 
     with open(filepath, "rb") as gzip_file:
         header = gzip_file.read(2)
-
     gzip_magic_number = b"\x1f\x8b"
     if header != gzip_magic_number:
         raise OSError(
             f"Input file {filepath} is not a compressed file.\n"
             + f"{filepath} has header of {header} and not gzip's {gzip_magic_number}.\n"
         )
+
     return gzip.open(filepath, "r")
 
 
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -122,36 +122,39 @@ def loads():
     pass
 
 
-def _extract_fileobj(fileobj):
+def _extract_fileobj(filepath):
     """
     Checks to see if a file is compressed, and if so, extract it with gzip
     so that the uncompressed file can be returned.
+    It returns a filename or file object containing XML data that will be
+    ingested by ``xml.etree.ElementTree.iterparse``.
 
     Args:
-        fileobj: A file path or a Python file object.
+        filepath: A path-like object or str.
 
     Returns:
-        pathlib.PosixPath or gzip.GzipFile: An uncompressed file object.
+        pathlib.PosixPath or gzip.GzipFile: A filename or file object containing
+          XML data.
     """
-    with open(fileobj, "rb") as gzip_file:
+    with open(filepath, "rb") as gzip_file:
         header = gzip_file.read(2)
     gzip_magic_number = b"\x1f\x8b"
 
-    return gzip.GzipFile(fileobj) if header == gzip_magic_number else fileobj
+    return gzip.GzipFile(filepath) if header == gzip_magic_number else filepath
 
 
-def readLHEInit(fileobj):
+def readLHEInit(filepath):
     """
     Read and return the init blocks. This encodes the weight group
     and related things according to https://arxiv.org/abs/1405.1067
 
     Args:
-        fileobj: A file path or a Python file object.
+        filepath: A path-like object or str.
 
     Returns:
         dict: Dictionary containing the init blocks of the LHE file.
     """
-    fileobj = _extract_fileobj(fileobj)
+    fileobj = _extract_fileobj(filepath)
     initDict = {}
     for event, element in ET.iterparse(fileobj, events=["end"]):
         if element.tag == "init":
@@ -189,8 +192,8 @@ def readLHEInit(fileobj):
     return initDict
 
 
-def readLHE(fileobj):
-    fileobj = _extract_fileobj(fileobj)
+def readLHE(filepath):
+    fileobj = _extract_fileobj(filepath)
     try:
         for event, element in ET.iterparse(fileobj, events=["end"]):
             if element.tag == "event":
@@ -206,12 +209,12 @@ def readLHE(fileobj):
         return
 
 
-def readLHEWithAttributes(fileobj):
+def readLHEWithAttributes(filepath):
     """
     Iterate through file, similar to readLHE but also set
     weights and attributes.
     """
-    fileobj = _extract_fileobj(fileobj)
+    fileobj = _extract_fileobj(filepath)
     try:
         for event, element in ET.iterparse(fileobj, events=["end"]):
             if element.tag == "event":
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -138,9 +138,7 @@ def _extract_fileobj(fileobj):
         header = gzip_file.read(2)
     gzip_magic_number = b"\x1f\x8b"
 
-    if header == gzip_magic_number:
-        fileobj = gzip.GzipFile(fileobj)
-    return fileobj
+    return gzip.GzipFile(fileobj) if header == gzip_magic_number else fileobj
 
 
 def readLHEInit(fileobj):
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -128,8 +128,7 @@ def _extract_fileobj(fileobj):
     so that the uncompressed file can be returned.
 
     Args:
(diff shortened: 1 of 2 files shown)


src/pylhe/__init__.py
```diff
@@ -120,14 +122,30 @@ def loads():
     pass
 
 
-def readLHEInit(thefile):
+def _open_gzip_file(filepath):
```
matthewfeickert, line 125: This is going to need work as [`xml.etree.ElementTree.iterparse`](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.iterparse) seems unhappy at the moment

```
$ file unweighted_events.lhe.gz 
unweighted_events.lhe.gz: gzip compressed data, was "tmp_0_unweighted_events.lhe", last modified: Tue Mar 30 07:22:36 2021, max compression, original size modulo 2^32 15580732
$ python
Python 3.8.6 (default, Jan  5 2021, 00:14:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pylhe
>>> pylhe.readLHEInit("unweighted_events.lhe.gz")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/feickert/Code/GitHub/Scikit-HEP/pylhe/src/pylhe/__init__.py", line 148, in readLHEInit
    for event, element in ET.iterparse(filepath, events=["end"]):
  File "/home/feickert/.pyenv/versions/3.8.6/lib/python3.8/xml/etree/ElementTree.py", line 1229, in iterator
    data = source.read(16 * 1024)
  File "/home/feickert/.pyenv/versions/3.8.6/lib/python3.8/gzip.py", line 288, in read
    self._check_not_closed()
  File "/home/feickert/.pyenv/versions/3.8.6/lib/python3.8/_compression.py", line 14, in _check_not_closed
    raise ValueError("I/O operation on closed file")
ValueError: I/O operation on closed file
```

src/pylhe/__init__.py
```diff
@@ -120,14 +122,30 @@ def loads():
...
+def readLHEInit(filepath):
     """
     Read the init blocks. Return dict. This encodes the weight group
     and related things according to https://arxiv.org/abs/1405.1067
     This function returns a dict.
     """
+    filepath = _open_gzip_file(filepath)
     initDict = {}
-    for event, element in ET.iterparse(thefile, events=["end"]):
+    for event, element in ET.iterparse(filepath, events=["end"]):
```
matthewfeickert, line 148: This seems to be a problem for unweighted events:

```
$ file unweighted_events.lhe.gz 
unweighted_events.lhe.gz: gzip compressed data, was "tmp_0_unweighted_events.lhe", last modified: Tue Mar 30 07:22:36 2021, max compression, original size modulo 2^32 15580732
$ python
Python 3.8.6 (default, Jan  5 2021, 00:14:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pylhe
>>> pylhe.readLHEInit("unweighted_events.lhe.gz")
weightgroup must have attribute 'type'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/feickert/Code/GitHub/Scikit-HEP/pylhe/src/pylhe/__init__.py", line 159, in readLHEInit
KeyError: 'type'
>>> pylhe.readLHEInit("unweighted_events.lhe")
weightgroup must have attribute 'type'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/feickert/Code/GitHub/Scikit-HEP/pylhe/src/pylhe/__init__.py", line 159, in readLHEInit
KeyError: 'type'
```

src/pylhe/__init__.py
```diff
@@ -120,14 +123,30 @@ def loads():
...
+def _open_gzip_file(filepath):
+    """
+    Checks to see if a file is compressed, and if so, open it with gzip
+    """
+    if not Path(filepath).name.lower().endswith(".gz"):
+        return filepath
+
+    with gzip.open(filepath, "r") as gzip_file:
+        try:
+            gzip_file.read(1)
```
matthewfeickert, line 135: Need to _actually_ check that this is a `.gz` file here and not just that it is possible to read here, as a totally blank file can be read with `gzip_file.read(1)` just fine.

lukasheinrich: I guess convention is fine.. we could check file headers.. but I feel like we could say that if you give pylhe a file endig w/ `.gz` it's your responsibility to ensure that it's  a gzipped file

matthewfeickert: > we could check file headers

Yeah, I think this is a better way to do it.

lukasheinrich: https://github.com/ahupp/python-magic this could be a library solution  or if we don'tt want tthe dependency read the fiirstt to bytes manually: https://stackoverflow.com/a/47080739/286382

matthewfeickert: I had tried reading the first bytes manually last night, but I was getting errors for files produced by `mg5`, but I must have done something wrong as this works fine

```
$ file unweighted_events.lhe
unweighted_events.lhe: exported SGML document, ASCII text
$ file unweighted_events.lhe.gz 
unweighted_events.lhe.gz: gzip compressed data, was "tmp_0_unweighted_events.lhe", last modified: Tue Mar 30 07:22:36 2021, max compression, original size modulo 2^32 15580732
$ python
Python 3.8.6 (default, Jan  5 2021, 00:14:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> with open("unweighted_events.lhe.gz", "rb") as readfile:
...     print(readfile.read(2))
... 
b'\x1f\x8b'

```

matthewfeickert: Oh it was because I was using `gzip.open` and not `open` to check:

```
>>> import gzip
>>> with gzip.open("unweighted_events.lhe.gz", "rb") as readfile:
...     print(readfile.read(2))
... 
b'<L'
```

src/pylhe/__init__.py
matthewfeickert: This isn't related to Issue #74

matthewfeickert:
@lukasheinrich I think this is finally ready for review (after me spamming this PR for the last hour). :P

src/pylhe/__init__.py
```diff
@@ -120,14 +123,34 @@ def loads():
     pass
 
 
-def readLHEInit(thefile):
+def _open_gzip_file(filepath):
```
lukasheinrich, line 126: this should probably rreturn a consistent type.. i.e. also a file object for the non-gz case.. i.e. `return open(filepath)`

if we're checkig the magic bytes anyways do we need the `gz` extension check?

src/pylhe/__init__.py
```diff
@@ -120,14 +123,34 @@ def loads():
...
+    return gzip.open(filepath, "r")
+
+
+def readLHEInit(filepath):
     """
     Read the init blocks. Return dict. This encodes the weight group
     and related things according to https://arxiv.org/abs/1405.1067
     This function returns a dict.
     """
+    filepath = _open_gzip_file(filepath)
```
lukasheinrich, line 151: variable should thenn prob be `fileobj` or similar

src/pylhe/__init__.py
```diff
@@ -179,13 +203,14 @@ def readLHE(thefile):
         return
 
 
-def readLHEWithAttributes(thefile):
+def readLHEWithAttributes(filepath):
```
lukasheinrich, line 206: variable should thenn prob be fileobj or similar

src/pylhe/__init__.py
matthewfeickert: > this should probably return a consistent type.. i.e. also a file object for the non-gz case.. i.e. `return open(filepath)`

Yeah, that's a good point, though that will need some refactoring with current use, as it will currently fail if you give it a non-gzip-ed LHE file then as `ET.iterparse` tries to open up the file you give it.

> if we're checking the magic bytes anyways do we need the `gz` extension check?

This was just so files that don't match the naming convention get passed through to `ET.iterparse` to be opened up. If you do match, then we check if you're actually a `gz` file.

matthewfeickert: So let me think on a refactor for that tonight and you can review again tomorrow. :+1:

src/pylhe/__init__.py
```diff
@@ -120,14 +123,45 @@ def loads():
...
+    if header != gzip_magic_number:
+        raise OSError(
+            f"Input file {fileobj} is not a compressed file.\n"
+            + f"{fileobj} has header of {header} and not gzip's {gzip_magic_number}.\n"
+        )
+
+    with gzip.open(fileobj, "rb") as readfile:
+        extracted_file = readfile.read()
+
+    return extracted_file
```
matthewfeickert, line 153: This is actually not good either. If you take the following madgraph script file:

```
$ cat drell-yan_test.mg5
generate p p > l+ l-
output drell-yan_test
launch drell-yan_test
```

and run it you'll have a file `drell-yan_test/Events/run_01/unweighted_events.lhe.gz`. If you take this script

```python
import gzip
from pathlib import Path

import pylhe

if __name__ == "__main__":
    unweighted_gzip = Path().cwd().joinpath("unweighted_events.lhe.gz")
    unweighted_file = Path().cwd().joinpath("unweighted_events.lhe")

    file_generator = pylhe.readLHE(unweighted_file)
    event = next(file_generator)
    print(f"uncompressed: {event.particles[-1].e}")

    file_generator = pylhe.readLHE(unweighted_gzip)
    event = next(file_generator)
    print(f"compressed:   {event.particles[-1].e}")
```

and run it you'll get

```
$ python issue_example.py 
uncompressed: 68.822088005
# and then a deluge of the contents of the LHE file being printed to screen
```

If these lines are replaced with

```python
    return gzip.open(fileobj, "rb")
```

then 

```
$ python example.py
uncompressed: 68.822088005
compressed:   68.822088005
```

So I probably need to look at what `ET.iterparse` is expecting if the goal is to have `_extract_gzip_file` return the same object type.

matthewfeickert: [`xml.etree.ElementTree.iterparse()`](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.iterparse) want a file object to be able to iterate on, so returning a `gzip.GzipFile` like

```python
return gzip.open(fileobj, "rb")
```

is actually the right type to return. However, isn't it bad to be returning with an `open`? Ideally this would be done with a context manager, but it isn't clear to me how to properly copy all the necessary information as

```python
with gzip.open(fileobj, "rb") as gzip_file:
    extracted_fileobj = gzip.GzipFile(
        filename=gzip_file.filename, fileobj=gzip_file.read()
    )
return extracted_fileobj
```

runs into problems later in `xml.etree.ElementTree.iterparse`.

matthewfeickert: ...and it seems I was just making it hard for myself as

```python
with gzip.open(fileobj, "rb") as gzip_file:
    gzip_fileobj = gzip.GzipFile(filename=gzip_file.filename)
return gzip_fileobj
```

works.

src/pylhe/__init__.py
matthewfeickert: c.f. my ranting below. :P

src/pylhe/__init__.py
```diff
@@ -120,14 +122,36 @@ def loads():
...
+        fileobj = gzip.GzipFile(fileobj)
+    return fileobj
+
+
+def readLHEInit(fileobj):
     """
     Read the init blocks. Return dict. This encodes the weight group
     and related things according to https://arxiv.org/abs/1405.1067
     This function returns a dict.
     """
```
lukasheinrich, line 151: add doctring .. is  fileobj a path or a literal object?

matthewfeickert: @lukasheinrich Can be either a path or a Python file object, I believe.

lukasheinrich: hm I would've thought the `open()` call in `_extract_fileobj` would forcce a path argument

matthewfeickert: We got rid of the `open` though so it is now just

```python
return gzip.GzipFile(fileobj) if header == gzip_magic_number else fileobj
```

You can see this more clearly in the tests that I added in https://github.com/scikit-hep/pylhe/pull/74/commits/13a60a10b9e747a6dcea313b99a39c508c614818

matthewfeickert: oh oh you mean the open in the check.

matthewfeickert: @lukasheinrich sorry I think I understand what you mean now &mdash; I was missing it before.

lukasheinrich (approved):

PR #75, opened 2021-03-31
docs: Revise author list

lukasheinrich:
cc @matthewfeickert 

```
* Add Matthew Feickert to author list in citation and Zenodo metadata
```

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -9,6 +9,11 @@
             "affiliation": "CERN",
             "name": "Lukas Heinrich",
             "orcid": "0000-0002-4048-7584"
+        },
+        {
+            "affiliation": "University of Illinois at Urbana-Champaign",
+            "name": "Matthew Feickert",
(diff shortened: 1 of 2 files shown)


matthewfeickert:
Oh, I should also add this to the Zenodo file.

matthewfeickert (approved):

eduardo-rodrigues:
Hi @matthewfeickert, for my knowledge, what is the `.zenodo.json` file used for? Something worth having in other packages? Advance thanks.

eduardo-rodrigues:
Thanks a lot @matthewfeickert. That's indeed nice to have the right names ones wants to display together with the affiliation and ORCID. What I like less is the fact that one needs to update the file for every release, and I'm not a fan of bumpversion, which you use. I may try a file with partial information, meaning a file without the version number, for example; I would expect Zenodo to fill in the blanks in the same way that it creates a JSON file on the fly from best guesses if no such file is present.

PR #87, opened 2021-04-19
refactor: Deprecate reader API for more Pythonic style

matthewfeickert:
Revise the reader API to be more Pythonic in style as opposed to more  C++/ROOT style. Example: `pylhe.readLHEWithAttributes` -> `pylhe.read_lhe_with_attributes`.

```
* Deprecate C style reader API for more Pythonic style
* Ignore flake8 warnings for E501: line too long
```


matthewfeickert:
I tried removing the CODECOV_TOKEN secret to avoid the problem of [the April 15th security problems](https://about.codecov.io/security-update/), but it seems that things aren't reporting now. :/

matthewfeickert:
> I tried removing the CODECOV_TOKEN secret to avoid the problem of [the April 15th security problems](https://about.codecov.io/security-update/), but it seems that things aren't reporting now. :/

Seems it is still required for uploading the coverage report:

> The token below is used exclusively for uploading coverage reports.

src/pylhe/__init__.py
```diff
@@ -312,3 +313,35 @@ def visualize(event, outputname):
     tex2pix.Renderer(tex).mkpdf(outputname)
     subprocess.check_call(["pdfcrop", outputname, outputname])
     os.remove("event.dot")
+
+
```
lukasheinrich, line 317: I think we can just do a hard rename.. it's a small enough package

matthewfeickert: I was thinking this would give the people we know are using the library at one release grace period. Though as Henry has mentioned in the past, no one really reads warnings. We could probably just tag everyone who has ever opened an Issue or contributed to `pylhe` after this PR is merged and then just announce to them here that we'll be breaking the API in the next release.

lukasheinrich (approved):

matthewfeickert:
@8me, @ariaradick, @shantanu-gontia, @eduardo-rodrigues, @fuenfundachtzig, @SundeMarius, @gridge

This is just a heads up that the next release of `pylhe` will be an API breaking change for the reader API.

PR #108, opened 2022-01-07
build: Add Python 3.10 to PyPI metadata

eduardo-rodrigues:
Indeed support for Python 3.10 was added in https://github.com/scikit-hep/pylhe/pull/105 but the setup.cfg qualifiers did not reflect that.

While at it I also moved to being more "demanding" on the `Particle` package dependency to ensure that PDG 2021 is used by default, which seems fine (I realise a new version is typically picked up but for formality I made the explicit upgrade).

```
* Add Python 3.10, Python 3 only, and Developers to classifiers
metadata for PyPI.
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ python_requires = >=3.6
 install_requires =
     networkx~=2.2
     tex2pix~=0.3
-    particle~=0.16
+    particle~=0.14
     awkward>=1.2.0
     vector>=0.8.1
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,14 +16,16 @@ classifiers =
     Development Status :: 4 - Beta
     License :: OSI Approved :: Apache Software License
     Intended Audience :: Science/Research
+    Intended Audience :: Developers
     Topic :: Scientific/Engineering
     Topic :: Scientific/Engineering :: Physics
     Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
-
+    Programming Language :: Python :: 3.10
 [options]
 package_dir =
     = src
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -33,7 +35,7 @@ python_requires = >=3.6
 install_requires =
     networkx~=2.2
     tex2pix~=0.3
-    particle~=0.14
+    particle~=0.16
```
matthewfeickert, line 38: Why does particle need to be updated? These should be lower bounds on the `pylhe` side.

matthewfeickert: I know you mentioned that you wanted to "ensure that PDG 2021", but we shouldn't need to update lower bounds on this, right? You haven't yanked older releases from PyPI as there isn't anything "wrong" with the older releases. I guess I'm not sure how to think about the versioning for `particle`. Should we think of the minor releases as being more like CalVer then SemVer?

eduardo-rodrigues: > Why does particle need to be updated? These should be lower bounds on the `pylhe` side.

Indeed you are largely shadowed from most developments in Particle since only dependent for functionality that is very stable. On the other hand Particle does evolve, with PDG updates, the odd fix, etc., and dependents should evolve accordingly IMO. For example at some point I also updated the dependency on Lark in Particle. But up to you. I will remove the change in the dependency since anyway PRs should be scoped and I'm doing 2 things here.

eduardo-rodrigues: Right, nothing wrong with older versions (modulo the odd buglet that is caught rarely but once in a while). No need to yank any release so far.

I do not use CalVer. I work more lile SemVer. So far there's never been a major 1.0 release because there are still some things I would like to freeze first in the API/package. As for minor version updates they mostly take place when moving to new data files (0.16 got out with the PDG 2021 data, for example); BTW, data files have a version too, just in case. Minor version changes also take place when fixes are needed to particles properties, as that kind of means a different behaviour since some methods get output changed (case when there was a bug in the parity of antibaryons). I hope this makes sense. (Probably more than you really wanted to ask :D.)

eduardo-rodrigues: This PR is hence ready.

matthewfeickert (approved):
LGTM. :+1: Thanks!

setup.cfg
matthewfeickert: Sorry I missed these! But thank you for the info, Eduardo!

Issue #63, opened 2020-12-17
Consider graphviz as alternative to pydot and others


matthewfeickert:
I use alternatives to get PDF/PNG/... files (if needed) in `DecayLanguage`: why not simply go for https://pypi.org/project/graphviz/ for dot files, which provide you with everything you need? To be honest I started by using `pydot` (still in use in `DecayLanguage`) but this is also seems unmaintained whereas `graphviz` is equivalent and very well maintained, hence I'm going to make the switch asap.

_Originally posted by @eduardo-rodrigues in https://github.com/scikit-hep/pylhe/issues/53#issuecomment-722022276_

Issue #53, opened 2020-11-04
Reconsider tex2pix dependency


matthewfeickert:
`tex2pix` is a [core dependency of `pylhe`](https://github.com/scikit-hep/pylhe/blob/d9ce20436161ec174bde6555c7dd6e68f7f5e0ea/setup.cfg#L40-L43) but the [last release of `v0.3.1` was in 2016](https://pypi.org/project/tex2pix/0.3.1/#history), so it seems to be safely "unmaintained" at this point. The source code also doesn't seem to be publicly available on GitHub.

:wave: @agbuckley, as the author of `tex2pix` (thanks! :bow:), can you comment on if there is any plans to maintain `tex2pix` in the future? Or would you recommend that we look for an alternative that is more probable to be patched if there are issues that are found. I can imagine that you already have more than enough on your plate and being responsible for maintaining another codebase might very understandably not be high on the list. :)

eduardo-rodrigues:
@matthewfeickert, allow me to chime in, since I use alternatives to get PDF/PNG/... files (if needed) in `DecayLanguage`: why not simply go for https://pypi.org/project/graphviz/ for dot files, which provide you with everything you need? To be honest I started by using `pydot` (still in use in `DecayLanguage`) but this is also seems unmaintained whereas `graphviz` is equivalent and very well maintained, hence I'm going to make the switch asap.

agbuckley:
Hi all,

I'm coming in cold on this. There hasn't been a new tex2pix release because I'm not aware of problems with it -- if there are, let me know! I will be keeping it working because pyslha uses it for spectrum plotting. But note that it was written mainly for pyslha and there are probably nicer ways to do what it does -- I just factored it off because it seemed useful.
As you noted, I'm pretty busy, so it's a backburner project like many of my TeX-related ones, especially during Covid when time is tighter than ever. This and several other projects were hosted on Bitbucket, but then they deleted their Mercurial support and I've not got round to converting the repos to git and pushing to GitLab yet... I'll put it on my todo list.
Andy

matthewfeickert:
> Hi all, I'm coming in cold on this.

Thanks. I didn't mean to drop you in without context, but I figured your insights would be useful here (they are).

> There hasn't been a new `tex2pix` release because I'm not aware of problems with it -- if there are, let me know!

Nope! Just wasn't sure what the state of things was if things did come up and where we would be able to open issues.

[...]

PR #118, opened 2022-04-21
feat: Add IPython visualisation and switch to graphviz

eduardo-rodrigues:
This closes https://github.com/scikit-hep/pylhe/issues/63 and also deals in some way with https://github.com/scikit-hep/pylhe/issues/53.

Examples usage:
snapshot1

Easy also to save avoiding any coupling of dependencies:
snapshot2

```
* Add graphviz>=0.12.0 as a core dependency.
* Add graphviz.Digrap graph object as .graph property to LHEFile.
   - Add support for IPython visualization.
* Add test for LHEEvent_graph.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI/CD
 
 on:
   push:
+    branches:
+      - master
   pull_request:
   # Run daily at 0:01 UTC
   schedule:
@@ -46,12 +48,16 @@ jobs:
       run: |
         pytest -r sx tests/
 
-    - name: Report coverage with Codecov
-      if: github.event_name == 'push' && matrix.python-version == '3.10' && matrix.os == 'ubuntu-latest'
+      # Report coverage for oldest and newest Python tested to guard against version differences
+    - name: Report core project coverage with Codecov
+      if: >-
+        github.event_name != 'schedule' &&
+        (matrix.python-version == '3.6' || matrix.python-version == '3.10') &&
+        matrix.os == 'ubuntu-latest'
       uses: codecov/codecov-action@v2
       with:
-        file: ./coverage.xml
-        flags: unittests
+        files: ./coverage.xml
+        flags: unittests-${{ matrix.python-version }}
         name: pylhe
 
   docker:
--- a/.github/workflows/semantic-pr-check.yml
+++ b/.github/workflows/semantic-pr-check.yml
@@ -1,0 +1,24 @@
+name: Semantic Pull Request
+
+on:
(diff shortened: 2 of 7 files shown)


eduardo-rodrigues:
A few comments at this point:

- Using `graphviz` makes it simple to visualise events in notebooks, as per the example above. That's why this PR closes https://github.com/scikit-hep/pylhe/issues/63.
- From playing around a bit I found it rather evolved the way to get a particular event, which needs for example `itertools.islice`. I would suggest a follow-up issue to write a little wrapper function that returns the n-th event.
- I'm here removing the method `LHEEvent.visualize(...)`. I find it that it couples unnecessarily and also provides a rather indirect way of visualiing the graphs since the method involves internally calls with `dot2tex` and `tex2pix`, and finally the actual visualisation of the produced PDF file. To me this is effectively obsoleted with the way I show in this PR for how to display on the fly or get the DOT file or save it to standard formats. For the same reason I would then also delete the standalone function `visualize(...)`, thereby simplifying dependencies and couplings in design.

Seems good enough for a v0. Happy to hear your thoughts :-).

kratsg (approved):

eduardo-rodrigues:
@lukasheinrich and/or @matthewfeickert, let me know what you think of this and also of removing the `visualize` function. I would be tempted to avoid a deprecated given that the user base is probably not gigantic. I would rather move towards a major release, which seems due (at least a minor release, if you think the present API is not quite OK for a 1.0). Thanks.

Note that the notebooks should then be refreshed and updated with more examples of functionality. I'm happy to help there, putting in this new functionality. Just let me know.

eduardo-rodrigues:
Hello @lukasheinrich and @matthewfeickert, shall we converge on this and make a new release?

matthewfeickert:
```python
import pylhe
from skhep_testdata import data_path
import itertools

lhe_file = data_path("pylhe-testfile-pr29.lhe")
events = pylhe.read_lhe_with_attributes(lhe_file)

print(f"Number of events: {pylhe.read_num_events(lhe_file)}")

event = next(itertools.islice(events, 1,2))
print(event.graph.source)

event.graph.render(filename="test", format="png")
```

is indeed a nice little example. :+1:

matthewfeickert (approved):
LGTM. :+1: Thanks @eduardo-rodrigues!

I'm going to write some short notes and then also release a new minor release given the addition of a new dependency. :+1:

eduardo-rodrigues:
Sounds good, thanks. As soon as I find a bit of free time I will add this example as a get started in the README, if that's OK. I do find these little helpers a good thing.

PR #130, opened 2022-06-08
docs: Add a 'get started' example to the README

eduardo-rodrigues:
Relates to task https://github.com/scikit-hep/pylhe/issues/30 on package documentation.

PR #131, opened 2022-06-08
build: Decouple and minimise dependencies for visualisation

### LINKED ISSUE #53: Reconsider tex2pix dependency

eduardo-rodrigues:
It is best to decouple as much as possible, in design, and with the dependency on Graphviz we now have for visualisation, we no longer need networkx, tex2pix and dot2tex. The package gets much lighter with no loss of functionality since, as I show in MR https://github.com/scikit-hep/pylhe/pull/130, it is trivial at present to visualize a graph directly in a notebook or simply save the graph in, say, PNG or PDF.

IMO this deals with https://github.com/scikit-hep/pylhe/issues/53.

Resolves #53

```
* Remove visualize method.
* Remove viz extra.
* Remove networkx and tex2pix dependencies in favor of using graphviz.
```


tests/test_visualize.py
```diff
@@ -1,21 +1,8 @@
...
 
-def test_visualize(tmpdir):
-    lhe_file = skhep_testdata.data_path("pylhe-testfile-pr29.lhe")
-    events = pylhe.read_lhe_with_attributes(lhe_file)
-
-    start_event = 1
-    stop_event = 2
-    filename = tmpdir.join(f"event{start_event}.pdf")
-    for event in itertools.islice(events, start_event, stop_event):
-        pylhe.visualize(event, filename)
```
matthewfeickert, line 16: Any objection to keeping this in to make sure that the graphs being produced are actually valid for `graph.render` but with

```python
event = next(itertools.islice(events, 1, 2))
event.graph.render(filename="test_event1", format="pdf", cleanup=True)
```

?

matthewfeickert (commented):
I think this is good, and solves a lot of problems. :rocket: I think the only thing that would be good to note/fix(?) for the future would be

that the old method allowed for LaTeX fonts


old_example

while using `graphviz.graphs.Digraph.render` doesn't seem to do that

test

Is there a way to get LaTeX fonts? This is somewhat steering back into the questions of Issue #53, which I haven't taken the time to review and refresh myself on, so apologies if all of this should be obvious.

tests/test_visualize.py
eduardo-rodrigues: That sounds good. You're basically suggesting to test what is exemplified in the README, and that only fair.

True that test coverage is not amazing yet and needs more work.

eduardo-rodrigues:
> I think the only thing that would be good to note/fix(?) for the future would be that the old method allowed for LaTeX fonts.

> Is there a way to get LaTeX fonts? This is somewhat steering back into the questions of Issue https://github.com/scikit-hep/pylhe/issues/53, which I haven't taken the time to review and refresh myself on, so apologies if all of this should be obvious.

It must be possible since both packages in the end write DOT language files. I used what I'm doing in DecayLanguage. I'm putting your request down as a to-do to look into this soon, after a few days off ... In the meantime, if this goes in we're good tp proceed with the other MRs dependent on this.

tests/test_visualize.py
eduardo-rodrigues: Now added via https://github.com/scikit-hep/pylhe/pull/131/commits/2c37af784231d618f05b0fe60d2a3909e65b5e0c.

eduardo-rodrigues:
Hello @matthewfeickert, I finally remembered to re-instate a test as you had before, see your comment https://github.com/scikit-hep/pylhe/pull/131#discussion_r901226407 above.

Coverage is to be worked on but that's not specific to this MR, rather a general issue that coverage is missing for certain classes, etc. That's clearly outsidethe scope of this MR. This is an important MR as it makes dependencies easier and unblocks some dependent MRs.

As for your comment https://github.com/scikit-hep/pylhe/pull/131#issuecomment-1162397488 I would gladly make this a follow-up issue as it is a quality of life improvement, and something I can look into in the future. Again, a different matter, second priority.

In short, I really think this is ready to do. If happy please approve and I will make the follow-up issue. Thanks a lot.

eduardo-rodrigues:
Follow-up issue created - https://github.com/scikit-hep/pylhe/issues/135 for discussion https://github.com/scikit-hep/pylhe/pull/131#issuecomment-1162397488.

eduardo-rodrigues:
Hi @matthewfeickert, I will merge this really shortly since the follow-up task is created and blocking this further is counter-productive and not helping anyone, certainly not the users. Coverage is a serious matter to deal with across the files, not just the 2-3 lines mentioned by codecov above ;-).

matthewfeickert:
> Hi @matthewfeickert, I will merge this really shortly since the follow-up task is created and blocking this further is counter-productive and not helping anyone, certainly not the users. Coverage is a serious matter to deal with across the files, not just the 2-3 lines mentioned by codecov above ;-).

Okay. :+1: I still have some concerns about trading one viz library for another that now changes how the visualizations look, but I'm willing to kick that down the road a bit to sometime when I have more time to do the work here for these concerns to matter.

PR #141, opened 2022-08-30
feat: more homogeneous usage of ParseError

eduardo-rodrigues:
I noticed that the `read_lhe` type of functions and others, which parse the XML, do not all catch parsing errors. This PR makes things more homogeneous. Note that

1. On purpose I am not trying here to deal with test coverage of these parsing errors. That's something to be done with a dedicated PR and will require a special LHE file, if not several.
2.  We should discuss at some point if parsing errors should raise errors or be dealt with smoothly, as now, with print statements. Again, I think it is better to make any modification to the way these errors are dealt with homogeneously across the various methods.

### DIFF
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -337,17 +337,21 @@ def read_lhe_with_attributes(filepath):
                         eventdict["attrib"],
                         eventdict["optional"],
                     )
-    except ET.ParseError:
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues:
This is ready given the scope, see description. Merging as I have tests in the pipeline to be added.

PR #146, opened 2022-09-16
test: first tests of awkward module

eduardo-rodrigues:


eduardo-rodrigues:
Hey @matthewfeickert, so this provides really simple tests for the awkward modulo. I feel we should have more tests on actual contents expected. Have you got suggestions?

I also wonder if the API is the best one, as one needs to go via somethig like `array.particles.vector` to get hold of the 4-momentum vectors. It should be just about trivial to be able to pick say 2 muons from an array sample, with adequate filtering, and plot their invariant mass. Again, thoughts? (This is beyond this PR but it made sense to write it down.)

eduardo-rodrigues:
Hey @matthewfeickert, as this PR is trivial and I know you are over-busy at the moment, let me bump up coverage and we come back to actual API discussions and more examples in the future :-). Seems reasonable. Merging. myself.

PR #148, opened 2022-10-07
docs: Add project logo

lukasheinrich:
@eduardo-rodrigues @matthewfeickert


eduardo-rodrigues:
Hey @lukasheinrich, nice :-).

Just a tiny comment - do you not want to add the actual file to a directory on the repository? It could be part of a docs folder to be used for Sphinx-based docs at some point.

eduardo-rodrigues (approved):

matthewfeickert:
Walking to office currently, but this also needs to be able to be properly loaded by PyPI so the path needs to be edited.

README.md
```diff
@@ -1,5 +1,8 @@
 # pylhe: Python LHE interface
 
+<img src="https://user-images.githubusercontent.com/2318083/194557871-b863d51a-b9b0-4884-9331-1ce37a195588.png" alt="drawing" width="250"/>
```
matthewfeickert, line 3: Ah okay, ignore my previous comment about this not rendering on PyPI as this is using the GitHub CDN and so would work. Though I agree with Eduardo that this would be good to have as a local file that could be used in a similar manner like https://github.com/scikit-hep/pyhf/blob/eac527e4601080df3c75db0276c208031f67e00c/README.rst?plain=1#L1-L4

README.md
```diff
@@ -1,5 +1,7 @@
 # pylhe: Python LHE interface
 
+<img src="https://raw.githubusercontent.com/scikit-hep/pylhe/master/docs/_static/img/pylhe-logo.png" alt="pylhe logo" width="250"/>
```
matthewfeickert, line 3: @lukasheinrich Is there any reason not to use SVG here? Given that we have it? 

pylhe-logo

matthewfeickert (approved):
We can always add a SVG later, so approving to get this in. Looks nice @lukasheinrich! :)

PR #153, opened 2022-10-25
ci: Enable Dependabot version updates for GitHub Actions

matthewfeickert:
* Add GitHub Dependabot config to check for GitHub Action version updates and to open PRs if available. Set the check and update interval to weekly.
   - c.f. https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues (approved):
Good to have this. Had not realised it was not in. I have it for my other packages and it is very useful. I set the interval to "monthly" but up to you.

PR #156, opened 2022-10-25
build: Support Python 3.7+

matthewfeickert:
```
* Require Python 3.7 for install.
* Drop Python 3.6 from testing in CI.
* Enforce Python 3.7+ syntax in pyupgrade.
```

Once Python 3.11 support is added (there is a small bug that can be fixed) and this is merged we could cut `v0.5.0` to get a clean Python version drop on a minor release.


eduardo-rodrigues (approved):
Good idea. We're "getting rid of" 3.6 across the org.

PR #180, opened 2023-01-23
fix: Update KeyError to new exception of DirectionalMap

tomeichlersmith:
The current code on `main` tries to catch a `KeyError` if a ID is not in the PDG to LaTeX name look up `DirectionalMap`, but the `DirectionalMap` throws a `MatchingIDNotFound` exception when it cannot find an ID.

This bug means that users cannot generate a graph with non-standard PDG IDs in them.

It is a simple patch to fix up this bug and allow for IDs not in the table to simple be represented by their ID number.

**Squash and merge commit message**

```
* Use MatchingIDNotFound exception over KeyError to allow for non-standard PDG
  IDs to be used.
* Add test for handling non-standard PDG IDs.
```

### DIFF
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -3,7 +3,8 @@ import xml.etree.ElementTree as ET
 
 import graphviz
 from particle import latex_to_html_name
-from particle.converters.bimap import DirectionalMaps, MatchingIDNotFound
+from particle.converters.bimap import DirectionalMaps
+from particle.exceptions import MatchingIDNotFound
 
 from pylhe._version import version as __version__
 from pylhe.awkward import register_awkward, to_awkward
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -3,7 +3,7 @@ import xml.etree.ElementTree as ET
 
 import graphviz
 from particle import latex_to_html_name
-from particle.converters.bimap import DirectionalMaps
+from particle.converters.bimap import DirectionalMaps, MatchingIDNotFound
 
 from pylhe._version import version as __version__
 from pylhe.awkward import register_awkward, to_awkward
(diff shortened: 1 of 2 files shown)


matthewfeickert (commented):
Thanks @tomeichlersmith. We should make sure that we have a test for this, and then cut a `v0.5.1`.

tomeichlersmith:
Alright, I've opened a PR to add a new test file: https://github.com/scikit-hep/scikit-hep-testdata/pull/112

I've also written a test that I think will fail because the test file isn't available yet.

matthewfeickert (approved):
LGTM. :+1: Thanks @tomeichlersmith!

matthewfeickert:
https://github.com/all-contributors please add @tomeichlersmith for code.

matthewfeickert:
@all-contributors please add @tomeichlersmith for code.

eduardo-rodrigues:
Hi @tomeichlersmith, good catch. For my knowledge which particle is 1023 in your events? I see such a number for the old Geant3 IDs (https://github.com/scikit-hep/particle/blob/master/src/particle/data/conversions.csv) but not for any PDG ID.

tomeichlersmith:
It's just a one-off ID my collaborators and I have been using for a "dark photon", kind of just an arbitrary choice to be honest.

eduardo-rodrigues:
OK, fair enough.

src/pylhe/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 import graphviz
 from particle import latex_to_html_name
-from particle.converters.bimap import DirectionalMaps
+from particle.converters.bimap import DirectionalMaps, MatchingIDNotFound
```
eduardo-rodrigues, line 6:
```suggestion
from particle.converters.bimap import DirectionalMaps
from particle.exceptions import MatchingIDNotFound
```

eduardo-rodrigues (approved):
I made a trivial correction of the import. Otherwise great. Thank you for this.

eduardo-rodrigues:
@tomeichlersmith, let us know if you have any suggestions for enhancements to the package and/or if you would like to contribute to it. For example, you might fancy the ability to retrieve extra bits of information from the file test file you submittted for the test added here, and we welcome that. Let us know and feel free to open an issue to discuss.

matthewfeickert:
@tomeichlersmith [`pylhe` `v0.5.1`](https://pypi.org/project/pylhe/0.5.1/) is now out on PyPI with this fix. :+1:

PR #184, opened 2023-02-03
chore: update to scikit-hep-testdata>=0.4.0

eduardo-rodrigues:
As per our recent chats.

While at it, update to require a more recent version of `scikit-hep-testdata` in tests. Indeed `scikit-hep-testdata` 0.4.0 is the first release that downloads files from the git repo on demand, rather than including them all in the PyPI package. Seems reasonable to be far more lightweight in the tests.

This should be the first step toward a release 0.6.0.


setup.cfg
```diff
@@ -35,8 +35,8 @@ python_requires = >=3.7
 install_requires =
     graphviz>=0.12.0
     particle>=0.16
-    awkward>=1.2.0
-    vector>=0.8.1
```
matthewfeickert, line 39: `vector` `v0.11.0` only requires `awkward>=1.2` for the `awkward` extra. 

https://github.com/scikit-hep/vector/blob/eb2b4142a28aacc3666197ba2c822b235e4c391b/pyproject.toml#L49-L51

Why do we need to bump the lower bound here? Is there something that breaks with `vector==0.8.1`?

matthewfeickert (commented):
Looks good @eduardo-rodrigues. Just one question RE: the `vector` bump &mdash; the rest make sense to me.

matthewfeickert:
The CI here is breaking because of the Poetry bug that @henryiii mentioned. @henryiii can you remind me what this fix is? Just update the `isort` release?

henryiii:
Yes, `pre-commit autoupdate` will do it.

matthewfeickert:
> Yes, `pre-commit autoupdate` will do it.

Thanks. This is done in PR #185 now.

setup.cfg
eduardo-rodrigues: There are fair points. To be honest I took the update from the scikit-hep metapackage where I ping =~ versions, and make consistent versions moving forward as the dependent packages get newer releases.

I realise the lower bounds ensure awkard v2, etc., are used. I thought I would make it explicit that we are moving to v2, hence this change. It's a different angle of thought from you.

This also explains the change in vector - not compulsory given the lower bound but it makes it clear that we take a version that supports awkward 2 (while keeping support for v1), see https://github.com/scikit-hep/vector/releases/tag/v0.11.0.

Up to you. I have no strong feelings. If you prefer we go with the trivial scikit-hep-testdata update and nothing else. Let me know.

eduardo-rodrigues: OK, I will reverse that part of this MR, then.

eduardo-rodrigues:
Merging since you @matthewfeickert effectively approved already.

matthewfeickert:
Thanks @eduardo-rodrigues! Should we make a `v0.6.0` release for this now? Or were there additional things you wanted to get in before this?

eduardo-rodrigues:
There are still issues to be tackled but I can't do them just now. We could do a 0.5.2 release since nothing but hooks and an update of scikit-hep-testdata's version got updated? Up to you; no strong feeling :-).

PR #195, opened 2023-04-14
feat: Auto register vector classes with awkward

### LINKED ISSUE #194: use scikit-hep.vector objects for momenta
I find myself often using the following syntax
```python
momentum = vector.Array(ak.flatten(array_from_lhe[<some selection>]['vector']))
```
so that I can use [vector](https://github.com/scikit-hep/vector)'s helper functions for calculating common variables.
I get the `array_from_lhe` by using `pylhe.to_awkward(pylhe.read_lhe(<some file>))`.

To me, from the outside, the existence of the `Momentum4D` record within the awkward array returned by `pylhe.to_awkward` seems to be extraneous especially since we have a module with a known awkward interface and designed for 4D vectors. Is replacing `Momentum4D` with vector's `Vector4D` something that is feasible? If this isn't feasible (or desirable) for some reason, this isn't too much boilerplate code to do what I want, I was just curious.

tomeichlersmith:
Both `vector` and `awkward` are required dependencies of `pylhe` at this time, so this PR proposes to make the registration of the vector classes that `pylhe` uses automatic. This means that the registration function would become a no-operation to maintain backwards compatibility and future scripts wouldn't have to worry about calling it.

Closes #194 .

## Questions
- [x] Is there a deprecation flag/decorator/documentation that I should add to that function?
- [x] ~Should I remove that registration function all-together? It would break the API, but it also is an easy thing to fix for users (just don't use that function anymore since it happens automatically).~ No
- [x] Add a test showing that the vector helpers work when accessed via a awkward array built by pylhe?

```
* Automatically register Awkward behaviors at pylhe import.
* Deprecate pylhe.awkward.register_awkward for removal in pylhe v0.8.0.
* Add test for Awkward behaviors registration.
* Add test for deprecation of pylhe.awkward.register_awkward.
* Update notebooks to use more Awkward behaviors.

Co-authored-by: Matthew Feickert <matthew.feickert@cern.ch>
```


src/pylhe/awkward.py
```diff
@@ -10,12 +10,22 @@ def __dir__():
...
+    import warnings
+
+    warnings.warn(
+        """The `pylhe.register_awkward` function does nothing.
+            The registration now happens automatically when `pylhe` is imported,
+            so calls to `pylhe.register_awkward` can be safely removed.""",
+        category=DeprecationWarning,
+        stacklevel=2,
+    )
+    pass
```
matthewfeickert, line 28: I'd prefer to use the system we have for `pyhf` of giving an explicit deprecated API warning with a removal version (2 minor releases). c.f.

* https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/src/pyhf/exceptions/__init__.py#L181-L190
* https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/src/pyhf/infer/intervals/__init__.py#L11-L30

No need to make an exceptions system here, I was just showing what the warning system looks like.

src/pylhe/awkward.py
```diff
@@ -58,3 +68,11 @@ class Event:
 
 class EventInfo:
     pass
+
+
+# register/bind our vector classes with awkward so
+# we can call thost helper functions
```
matthewfeickert, line 74:
```suggestion
# Register Awkward behaviors
```

tests/test_api.py
```diff
@@ -32,6 +32,9 @@ def test_top_level_api():
 def test_awkward_api():
     assert dir(pylhe.awkward) == ["register_awkward", "to_awkward"]
 
+    with pytest.warns(DeprecationWarning):
+        pylhe.register_awkward()
```
matthewfeickert, line 36: This should be its own test. c.f. https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/tests/test_regression.py#L183-L196 as an example, but doesn't need to be as detailed as that test.

tests/test_awkward.py
```diff
@@ -13,3 +13,7 @@ def test_to_awkward():
     arr = pylhe.to_awkward(pylhe.read_lhe_with_attributes(TEST_FILE))
     assert len(arr) == 791
     assert len(arr.eventinfo.nparticles) == len(arr.particles)
+    # iss 194 pr 195 - make sure vector helper classes funciton properly
+    #    here we make sure the 'mass' helper function on 4D vectors
+    #    is called on each particle's momentum
```
matthewfeickert, line 18:
```suggestion
```

This commentary is good for the PR body, but isn't really helpful in the tests. If test aren't clear what they're doing then they should get a docstring, not comments.

matthewfeickert (changes requested):
Thanks for your PR @tomeichlersmith. This looks pretty good, and I just have a few small suggestions. :+1:

tests/test_awkward.py
```diff
@@ -13,3 +13,13 @@ def test_to_awkward():
     arr = pylhe.to_awkward(pylhe.read_lhe_with_attributes(TEST_FILE))
     assert len(arr) == 791
     assert len(arr.eventinfo.nparticles) == len(arr.particles)
+
+
+def test_awkward_registration():
+    """awkward registration should correctly happen on import
```
eduardo-rodrigues, line 19:
```suggestion
    """Awkward registration should correctly happen on import.
```

tests/test_awkward.py
```diff
@@ -13,3 +13,13 @@ def test_to_awkward():
...
+
+    make sure vector helper classes function properly by calling
```
eduardo-rodrigues, line 21:
```suggestion
    Make sure vector helper classes function properly by calling
```

eduardo-rodrigues:
This sounds very good. @tomeichlersmith, would it make sense to enhance the notebook example a little bit to show how easier it is now to perform calculations such as invariant masser or cone distances or ...?

Otherwise I'm ready to approve.

Many thanks.

tomeichlersmith:
The `awkward_example` jupyter notebook already uses the vector method `mass` to calculate the invariant mass of two particles, but I could look at updating `zpeak`. Where is the `unweighted_events.lhe` file that is referenced in that example?

eduardo-rodrigues:
Indeed - I should have checked.

The zpeak notebook needs to be updated since the LHE file it mentions is not in the package nor in scikit-hep-testdata, I recall. It would deserve a new test file ... In that respect, BTW, since you say you regularly use pylhe, would you happen to have interesting (small) files to enhance the test suite in scikit-hep-testdata? We are always keen on having better coverage.

eduardo-rodrigues (approved):

tomeichlersmith:
I just used the drell-yan test example for the zpeak and it appears to be the same file (the first two events are the same graphs as before).

I've already included my small and interesting file :smile: It was needed for PR #180 to show that a unregistered PDG ID can exist in the LHE file and the graph can still be rendered.

eduardo-rodrigues:
Ah ha, thanks, I knew I had exchanged with you already in the past - too many repos and pull requests ;-).

eduardo-rodrigues (approved):
This seems ready to me. On to @matthewfeickert. Approving.

src/pylhe/awkward.py
```diff
@@ -10,12 +10,28 @@ def __dir__():
 
 
 def register_awkward():
-    """Register Awkward behaviors for pylhe."""
+    """This function does nothing and is merely here for backwards compatibility
+    of the pylhe API
```
matthewfeickert, line 14:
```suggestion
    """
```

src/pylhe/awkward.py
```diff
@@ -10,12 +10,28 @@ def __dir__():
...
-    ak.mixin_class(ak.behavior)(Particle)
-    ak.mixin_class(ak.behavior)(Event)
-    ak.mixin_class(ak.behavior)(EventInfo)
+    The registration is done automatically when pylhe is imported since awkward
+    is a required dependency of pylhe.
+
+    .. deprecated:: 0.5.3
+       Simply remove calls since the registration happens automatically now.
+    .. warning:: :func:`~pylhe.register_awkward` will be removed in
+     ``pylhe`` ``v0.7.0``.
```
matthewfeickert, line 22:
```suggestion
    .. deprecated:: 0.6.0
       Remove use of :func:`~pylhe.register_awkward` as registration is automatic.
    .. warning:: :func:`~pylhe.register_awkward` will be removed in
     ``pylhe`` ``v0.8.0``.
```

This is a change in behavior and so not something that should go into a patch release (also `pylhe` is currently `v0.5.1` not `v0.5.2`).

src/pylhe/awkward.py
```diff
@@ -10,12 +10,28 @@ def __dir__():
...
+    """
+    import warnings
+
+    warnings.warn(
+        """The `pylhe.register_awkward` function does nothing and has been deprecated
+            as of pylhe v0.5.3 and will be removed in pylhe v0.7.0.
+            The registration now happens automatically when `pylhe` is imported,
+            so calls to `pylhe.register_awkward` can be safely removed.""",
```
matthewfeickert, line 30:
```suggestion
"pylhe.register_awkward is deprecated as of pylhe v0.6.0 and will be removed in pylhe v0.8.0."
 + " Please remove use of pylhe.register_awkward in favor of automatic registration."
```

No one reads warnings realistically, so if they are going to be read they should be short.

src/pylhe/awkward.py
```diff
@@ -10,12 +10,28 @@ def __dir__():
...
+        category=DeprecationWarning,
+        stacklevel=3,
+    )
+    pass
```
matthewfeickert, line 34:
```suggestion
```

Don't use `pass`.

tests/test_awkward.py
```diff
@@ -13,3 +13,13 @@ def test_to_awkward():
...
     assert len(arr) == 791
     assert len(arr.eventinfo.nparticles) == len(arr.particles)
+
+
+def test_awkward_registration():
+    """Awkward registration should correctly happen on import.
+
+    Make sure vector helper classes function properly by calling
+    a property provided by vector `mass` on each particle's momentum.
+    """
```
matthewfeickert, line 23:
```suggestion
```

My point earlier was that _if_ a test needs comments then they should go in the docstring. This is a simple test and the name alone makes it clear what is happening. You wrote a good test, so this is fine as is. :+1:

matthewfeickert (commented):
This all LGTM basically. I'm just going to take it the rest of the way and simplify somethings and try to reducde the noise of the notebook changes (we should probably move to using `jupytext` anyway).

Thanks for the PR @tomeichlersmith!

src/pylhe/awkward.py
```diff
@@ -10,12 +10,21 @@ def __dir__():
...
 
-    vector.register_awkward()
-    ak.mixin_class(ak.behavior)(Particle)
-    ak.mixin_class(ak.behavior)(Event)
-    ak.mixin_class(ak.behavior)(EventInfo)
+    warnings.warn(
+        "pylhe.awkward.register_awkward is deprecated as of pylhe v0.6.0 and will be removed in pylhe v0.8.0."
+        + " Please remove use of pylhe.awkward.register_awkward in favor of automatic registration.",
+        category=DeprecationWarning,
+        stacklevel=2,  # Raise to user level
```
matthewfeickert, line 26: @tomeichlersmith just to explain my changes, this should be at the level where the user will hit in when running a script in their logs easily / in the REPL.

matthewfeickert (approved):

matthewfeickert:
@eduardo-rodrigues Are you fine with me releasing `pylhe` `v0.6.0` with this change in it?

eduardo-rodrigues:
Totally. You are right, it does make more sense to do a 0.6.0 release given the small change in API.

PR #232, opened 2024-02-20
fix: dict classes should produce instances of themselves

APN-Pucky:
`LHEInit` and `LHEProcInfo` inherit from `dict`. When parsing a string to either of them with `fromstring` it should return the corresponding class not a generic dict. They should have the same `__init__` function as a `dict` for convenience.

```
* `LHEInit` and `LHEProcInfo` inherit from `dict`. When parsing a string
  to either of them with `fromstring` it should return the corresponding
  class not a generic dict. They should have the same `__init__` function
  as a `dict` for convenience.
```

### DIFF
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -185,12 +185,12 @@ class LHEInit(dict):
         "numProcesses",
     ]
 
-    def __init__(self):
-        pass
+    def __init__(self,*args,**kwargs):
+        super().__init__(*args,**kwargs)
 
     @classmethod
     def fromstring(cls, string):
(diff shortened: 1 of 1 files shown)


matthewfeickert (approved):
Thanks @APN-Pucky. This LGTM and I'd like this to go in before PR #233 so that things can be more atomic.

Can you please write a summary in the PR body of why this is needed, so that this can also get included in the squash and merge commit message?

APN-Pucky:
> Can you please write a summary in the PR body of why this is needed, so that this can also get included in the squash and merge commit message?

Done

Issue #251, opened 2024-08-26
Add documentation for event writing


matthewfeickert:
> My only comment to you @APN-Pucky is that the functionality should be visible as an example notebook for sure, probably in a dedicated new notebook (otherwise many will not realise writing is now available). Likewise I would add a little paragraph/sentence in the README.

_Originally posted by @eduardo-rodrigues in https://github.com/scikit-hep/pylhe/issues/233#issuecomment-2309825784_

Issue #177, opened 2023-01-04
Write `developer.md` to explain maintainers process


matthewfeickert:
Following up on https://github.com/scikit-hep/pylhe/pull/157#issuecomment-1371473774 write up notes that explain how to take care of releases and general maintenance.

matthewfeickert:
@eduardo-rodrigues while I still need to make a maintainers doc I recorded a video of me releasing `pylhe` `v0.5.1` (and trying to explain what I'm doing):

https://youtu.be/tsAWDumpcW8

PR #256, opened 2024-09-22
docs: LHE file writing examples

APN-Pucky:
This adds two notebooks one showing how to write custom LHE files with a simple Monte Carlo event generator example and one filtering out events with a Z boson in the event record.

Further, the previously empty `class LHEFile` now contains two fields: `info` and `events` with a write function to a output stream. Contrary to the previous conversion to a string and then writing the string to the file, this approach allows us to write large filtered files directly per event to disk without consuming much memory.

One minor source for confusion might still remain: `write_lhe_file` takes a `init` and `events` while `read_lhe_file` returns a `LHEFile`. I'd change `write_lhe_file`, introduced in recent 0.9.0, to only require an LHEFile. The same then applies to `write_lhe_string`.

Closes: #251


eduardo-rodrigues:
Hello @APN-Pucky. This  sounds great. I will try and have a proper look later in the week.

May I suggest that you also update the README with a paragraph or so about writing? One could even explicitly mention the existence of notebooks for how to get started with examples.

README.md
```diff
@@ -59,6 +61,78 @@ event.graph.render(filename="test", format="png", cleanup=True)
 event.graph.render(filename="test", format="pdf", cleanup=True)
 ```
 
+### Writing
+
+For a full example see [write](examples/write_monte_carlo_example.ipynb) or [filter](examples/filter_events_example.ipynb).
+The values in the sketch below are intentionally left empty since they depend on the use-case.
+The data structure of pylhe is:
```
eduardo-rodrigues, line 68:
```suggestion
The data structure of `pylhe` is:
```

src/pylhe/__init__.py
```diff
@@ -369,6 +369,42 @@ def fromstring(cls, string):
...
+        output_stream.write(
+            f"<LesHouchesEvents version=\"{self.init['LHEVersion']}\">\n"
+        )
+        output_stream.write(self.init.tolhe() + "\n")
+        for e in self.events:
+            output_stream.write(e.tolhe(rwgt=rwgt, weights=weights) + "\n")
+        output_stream.write("</LesHouchesEvents>")
+        return output_stream
+
+    def tolhe(self, rwgt=True, weights=False) -> str:
```
eduardo-rodrigues, line 390: I'm a bit on the fence, here. Given the way we write methods I would be tempted to have this as `to_lhe`.

eduardo-rodrigues: I see. OK, let's leave it, then.

eduardo-rodrigues:
Hi @APN-Pucky, I reckon you can go ahead and then release a new version so that we have support for 3.13 out there :-). Advance thanks.

APN-Pucky:
Good I'll try to follow https://www.youtube.com/watch?v=tsAWDumpcW8 and write a developer.md for #177 afterwards.

PR #261, opened 2024-10-07
add: APN to dependabot reviewers

APN-Pucky:


eduardo-rodrigues (approved):
Sure, good catch :-).

eduardo-rodrigues:
You left in draft mode but I guess this is it. Undraft if you agree and I will merge.

PR #262, opened 2024-10-07
chore: add 3.13 classifier and test

henryiii:
Adding 3.13, which was released today.


APN-Pucky:
@henryiii Do you think `lint.yml` and `publish-package.yml` should also already change to `python-version: '3.13'`?

APN-Pucky (approved):

henryiii:
It would be good to get patch release in the near future with the classifier since 3.13 is out! ;)

PR #263, opened 2024-10-07
chore: use ruff instead of flake8,black,isort,pyupgrade,nbQA,etc.

### LINKED ISSUE #247: Change formatting to use Ruff
https://github.com/astral-sh/ruff

henryiii:
Simpler, faster, less to manage.

Close #247.


src/pylhe/__init__.py
```diff
@@ -238,13 +234,12 @@ def _indent(elem, level=0):
             elem.text = i + "  "
         if not elem.tail or not elem.tail.strip():
             elem.tail = i
-        for elem in elem:
-            _indent(elem, level + 1)
+        for inner_elem in elem:
+            _indent(inner_elem, level + 1)
         if not elem.tail or not elem.tail.strip():
```
henryiii, line 239: Wait, having the later elem reference the inner elem is intentional? It's part of the tests, anyway.

APN-Pucky: I should check that carefully. it could be that the test is copied from pylhe's output to avoid breaking it.

APN-Pucky: But since I stole it from https://stackoverflow.com/a/33956544 with support from old pythons, I think it works as desired (given the 95 up votes...).

APN-Pucky: IMO one can just rewrite the later uses of `elem` to `inner_elem` too. Then it is at least clear that it is overridden.

henryiii: I should be able to make this explicit, let's see. It's hard to tell from the diff which is better.

henryiii: 3.8 was locked and frozen as EoL today, by the way.

henryiii:
FYI, I might not have included the absolute import check, it's somewhere in Ruff, but I don't use it, so not sure where it is.

src/pylhe/__init__.py
henryiii:
```python
for inner_elem in elem:
    _indent(inner_elem, level + 1)
if not elem[-1].tail or not elem[-1].tail.strip():
    elem[-1].tail = i
```
is better, IMO.

henryiii: FYI, the logic itself is correct, though, you do want to handle the last one like this.

matthewfeickert:
@henryiii thanks for this PR! I'm in transit but I can also review this later on this week, though seems @APN-Pucky is already in it. 👍

APN-Pucky (approved):

APN-Pucky:
LGTM, Thanks!

eduardo-rodrigues:
(Just added a label - we try and keep it tidy :).)

Issue #281, opened 2025-02-03
Bug in fieldnames list


matthewfeickert:
Woops. Looks like there's a bug here

https://github.com/scikit-hep/pylhe/blob/7519967c13c97a3e2c0a8918e2191040ddb34a3e/src/pylhe/__init__.py#L251

as of PR https://github.com/scikit-hep/pylhe/pull/233.

We should add a test to catch this and make a patch release.

cc @APN-Pucky @eduardo-rodrigues

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pylhe/pull/280#discussion_r1940239265_

APN-Pucky:
@matthewfeickert Yes, it should be fixed, but does it have any implications? I think not since tests passed using the `self["weightgroup"]` dict way. I know that the `__all__` variable is tested somehow (either pre-commit or ci).

PR #233, opened 2024-02-20
feature: add write lhe file functions

APN-Pucky:
I decided to add a `tolhe()` function to every class (since it is neither xml nor a simple string).
The format strings can maybe still be improved?
`read_lhe_init()` now returns a `LHEInit` object combining weights, procinfo and initinfo. for backwards compatibility I added the get/set there so that `LHEInit` will still behave as previous `LHEInit`, what is now `LHEInitInfo`.

The tests can probably still be extended, but for the first draft I just hardcoded them for one file.

Already includes https://github.com/scikit-hep/pylhe/pull/232 and https://github.com/scikit-hep/pylhe/pull/231
Closes: #229

PR #282, opened 2025-02-03
fix: Add missing comma in LHEInit fieldnames

matthewfeickert:
Fixes the bug in Issue #281 

Not sure we have a good test at the moment though.

### DIFF
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -248,7 +248,7 @@ def _indent(elem, level=0):
 class LHEInit(dict):
     """Store the <init> block as dict."""
 
-    fieldnames = ["initInfo", "procInfo" "weightgroup", "LHEVersion"]
+    fieldnames = ["initInfo", "procInfo", "weightgroup", "LHEVersion"]
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)


APN-Pucky:
Here's what we could do. But it seems a bit unnecessary unless we want to restrict the access to the predefined fields (breaking previous "free" behavior). And just defining the list of fields in a second place (tests) seems a bit wasted/stupid as a test. 
```py
class CustomDict(dict):
    fieldnames = ['name', 'age', 'email']

    def __setitem__(self, key, value):
        if key not in self.fieldnames:
            raise KeyError(f"Invalid key: {key}. Allowed keys are {self.fieldnames}")
        super().__setitem__(key, value)
```

```py
import pytest
from your_module import CustomDict  # Import your class

def test_valid_keys():
    data = CustomDict()
    data['name'] = 'Alice'  # Should work
    data['age'] = 30        # Should work

def test_invalid_key():
    data = CustomDict()
    with pytest.raises(KeyError, match="Invalid key: address"):
        data['address'] = '123 Street'  # Should fail
```

eduardo-rodrigues:
Hello. I tend to agree  that a test seems a bit of an overkill. On the other hand I'm sad the bug did not get caught.

matthewfeickert:
> I must say I don't see the point in having the variable fieldnames, unless we loop over it, but maybe there are some builtin `dict` functions that use it.

> I tend to agree that a test seems a bit of an overkill.

Yeah, I agree with you both that adding a test as is seems like a lot of superfluous work. I think we can skip that for this PR and maybe just call this good? 

Though I agree with @APN-Pucky that I don't off the top of my head remember why `LHEInit` has `fieldnames` 

https://github.com/scikit-hep/pylhe/blob/7519967c13c97a3e2c0a8918e2191040ddb34a3e/src/pylhe/__init__.py#L251

as `LHEInit` doesn't have a `fromstring` like `LHEInitInfo` does

https://github.com/scikit-hep/pylhe/blob/7519967c13c97a3e2c0a8918e2191040ddb34a3e/src/pylhe/__init__.py#L340-L342

and so `fieldnames` is never used in `LHEInit` (which is why we've never seen this as a bug).

Ah I guess looking a bit more the `fieldnames` were added in PR https://github.com/scikit-hep/pylhe/pull/233 when large components of what used to be `LHEInit` got moved into `LHEInitInfo`.

@APN-Pucky @eduardo-rodrigues If we don't explicitly use 

https://github.com/scikit-hep/pylhe/blob/7519967c13c97a3e2c0a8918e2191040ddb34a3e/src/pylhe/__init__.py#L251

then should we keep them at all? (I am going through this quickly while task switching, so I might be missing the obvious.)

src/pylhe/__init__.py
```diff
@@ -248,7 +248,7 @@ def _indent(elem, level=0):
 class LHEInit(dict):
     """Store the <init> block as dict."""
 
-    fieldnames = ["initInfo", "procInfo" "weightgroup", "LHEVersion"]
+    fieldnames = ["initInfo", "procInfo", "weightgroup", "LHEVersion"]
```
matthewfeickert, line 251:
```suggestion
```

> If we don't explicitly use
> 
> https://github.com/scikit-hep/pylhe/blob/7519967c13c97a3e2c0a8918e2191040ddb34a3e/src/pylhe/__init__.py#L251
> 
> then should we keep them at all?

Given that this line was never getting used (else we would have seen this non-list list break) I think we can just remove this.

eduardo-rodrigues: I had a look at the module. Most other classes have a fromstring method, which uses fieldnames. This one LHEInit doesn't. I'm no expert to understand why. Would it be useful to add such a method?

In any case, one thing the variable fieldnames is good for is documentation. Right now the docstring is minimalistic and maybe that's worth a revamp as well?

APN-Pucky: > Most other classes have a fromstring method, which uses fieldnames. This one LHEInit doesn't. I'm no expert to understand why. Would it be useful to add such a method?

I would be in favor of adding a `fromstring` there too. It sort of exists as part of `read_lhe_init(file_or_string)` function, but that can then be mapped to use the new `fromstring`.

APN-Pucky: I created a PR to add fromstring https://github.com/scikit-hep/pylhe/pull/283, but just like the `read_lhe_init` it does not use `fieldnames` and I don't see the point in forcing it to use it now.

APN-Pucky (approved):

eduardo-rodrigues (approved):
Let me approve also. Let's get this bug fix in asap and then continue with the other bigger update and improvement from you @APN-Pucky.

(I won't be responsive in the next few days.)

PR #283, opened 2025-02-07
feat: LHEInit.fromstring

APN-Pucky:


APN-Pucky:
I am not sure why the coverage decreases... The functions are the same and I added even more to the tests.

eduardo-rodrigues:
Hi @APN-Pucky, it seems it is because of the extra "branching" introduced with exceptions and if-then's. Not a huge difference. Dunnot if that's trivial for you to add some little tests similar to the one you added, where you provoke an exception, for example?

Otherwise this seems ready, nice 👍.

(I will be slow for the remainder of the week, at a workshop.)

src/pylhe/__init__.py
```diff
@@ -366,6 +315,106 @@ def fromstring(cls, string):
...
+
+                        initDict["weightgroup"][wg_type] = _temp
+            if element.tag == "LesHouchesEvents":
+                initDict["LHEVersion"] = float(element.attrib["version"])
+            if element.tag == "event":
+                break
+        return cls(**initDict)
+
+    @classmethod
+    def fromstring(cls, string):
```
eduardo-rodrigues, line 414:
```suggestion
    def fromstring(cls, string):
        """Create an instance from a string."""
```

tests/test_lhe_reader.py
```diff
@@ -218,3 +218,54 @@ def test_issue_102():
     assert len(list(pylhe.read_lhe(TEST_FILE_LHE_v3))) == len(
         list(pylhe.read_lhe_with_attributes(TEST_FILE_LHE_v3))
     )
+
+
+def test_read_lhe_init_raises():
+    """
+    Test that the init is read correctly.
```
eduardo-rodrigues:
```suggestion
    Test that the <init> block is read correctly.
```

src/pylhe/__init__.py
```diff
@@ -366,6 +315,107 @@ def fromstring(cls, string):
...
+                for child in element:
+                    # Find all weightgroups
+                    if child.tag == "weightgroup" and child.attrib != {}:
+                        if "type" in child.attrib:
+                            wg_type = child.attrib["type"]
+                        elif "name" in child.attrib:
+                            wg_type = child.attrib["name"]
+                        else:
+                            ve = "weightgroup must have attribute 'type' or 'name'."
+                            raise ValueError(ve)
```
eduardo-rodrigues, line 389: Wondering, would it be better to raise an AttributeError here and below?

tests/test_lhe_reader.py
```diff
@@ -218,3 +218,54 @@ def test_issue_102():
     assert len(list(pylhe.read_lhe(TEST_FILE_LHE_v3))) == len(
         list(pylhe.read_lhe_with_attributes(TEST_FILE_LHE_v3))
     )
+
+
+def test_read_lhe_init_raises():
+    """
+    Test that the <init> block is read correctly.
```
eduardo-rodrigues, line 225: Amend the doctring to state that the test raises an exception for missing block attributes.

It seems better to have the 2 tests separately, meaning the 2 raise. It's a minor thing, I know.

eduardo-rodrigues (approved):
I left trivial things and reckon that this nice addition is otherwise ready, hence am approving.

Thanks a lot!

APN-Pucky:
Thanks

Issue #295, opened 2025-06-19
awkward representation of momenta should not be vector but momentum


APN-Pucky:
While reviewing https://github.com/scikit-hep/pylhe/pull/294 I noticed that the momenta are stored as vectors in awkward and not momenta, which is confusing because `x` is used to represent `px` and `t` for `E`.

However the `vector` packages seems to fully support `Momenta4d` (by now) as a type.

APN-Pucky:
This mapping is what I am talking about https://github.com/scikit-hep/pylhe/blob/d9bda84af5b881d830f3d5b761783472e3ba45f5/src/pylhe/awkward.py#L27C1-L32C30

ikrommyd:
They are all aliased in vector though. You can name them with any of the compatible names
```
In [5]: events.particles.vector.typestr
Out[5]: '10000 * var * Momentum4D[x: float64, y: float64, z: float64, t: float64]'

In [6]: events.particles.vector.x
Out[6]: <Array [[0, -0, -379, ..., 622, -90.4, -288], ...] type='10000 * var * float64'>

In [7]: events.particles.vector.px
Out[7]: <Array [[0, -0, -379, ..., 622, -90.4, -288], ...] type='10000 * var * float64'>

In [8]: events.particles.vector.y
Out[8]: <Array [[0, -0, -659, ..., 483, -213, -446], ...] type='10000 * var * float64'>

In [9]: events.particles.vector.py
Out[9]: <Array [[0, -0, -659, ..., 483, -213, -446], ...] type='10000 * var * float64'>

In [10]: events.particles.vector.z
Out[10]: <Array [[348, -2.66e+03, ..., -377, -587], ...] type='10000 * var * float64'>

In [11]: events.particles.vector.pz
Out[11]: <Array [[348, -2.66e+03, ..., -377, -587], ...] type='10000 * var * float64'>

In [12]: events.particles.vector.t
Out[12]: <Array [[348, 2.66e+03, ..., 442, 792], ...] type='10000 * var * float64'>

In [13]: events.particles.vector.E
Out[13]: <Array [[348, 2.66e+03, ..., 442, 792], ...] type='10000 * var * float64'>

In [14]: events.particles.vector.energy
Out[14]: <Array [[348, 2.66e+03, ..., 442, 792], ...] type='10000 * var * float64'>
```

ikrommyd:
[...]

PR #294, opened 2025-06-17
feat: add `weights` to the awkward array if they are present

ikrommyd:

PR #296, opened 2025-06-19
fix: drop spatial_momentum_map

APN-Pucky:
Closes: #295 

- [x] Add some tests that check the vectors values both in `x` and `px`.
- [x] Wait for https://github.com/scikit-hep/pylhe/pull/294 because `TEST_FILE` is renamed.


tests/test_awkward.py
```diff
@@ -58,3 +59,22 @@ def test_awkward_registration():
     arr = pylhe.to_awkward(pylhe.read_lhe_with_attributes(TEST_FILE_WITH_WEIGHTS))
     assert len(arr.particles.vector.mass) == len(arr.particles)
     assert len(arr.particles.vector.mass) == len(arr.weights)
+
+
+def test_to_awkward_vector():
+    """
+    Test numeric equality of momentas represented by vector.
```
eduardo-rodrigues, line 66:
```suggestion
    Test numeric equality of momenta represented by vectors.
```

APN-Pucky: Thanks. Since this was unlikely to be the only spelling error I added the codespell precommit hook and fixed the errors it showed.

PR #312, opened 2025-09-02
docs: docstring style from markdown to rst

APN-Pucky:


eduardo-rodrigues:
Thanks - not always easier to get it right between the 2 ;-).

PR #321, opened 2025-10-22
docs: Binder PyHEP 2025 talk

APN-Pucky:
Test URL (Readme link is already for main branch): https://mybinder.org/v2/gh/scikit-hep/pylhe/binder_talk?urlpath=lab/tree/docs/source/examples/00_version_1.0.0.ipynb

Talk:
- explain LHE format a bit as introduction
- give a quick easy entry point via `LHEFile`
- what is new in version 1.0.0
- focus on typing and dict -> dataclass transition
- link to existing examples

As a side product there are also several pre-1.0.0 improvements:

- `LHEEvent` `Optional[dict/list]` was redundant since they can just remain
  empty in the dataclass
- `fromstring`/`frombuffer`/`fromfile` for `LHEFile`
- `fromstring`/`frombuffer` for `LHEEvent` that returns
  `Generator`/`Iterable[LHEEvent]`
- deprecation of `read_lhe`, `read_lhe_file` and `read_lhe_with_attributes`
  in favour of above `from*` functions, which streamline the inclusion of
  attributes (`with_attributes`) to a single function
- Recycle opened files/buffers through `buffer.seek(0) `instead of
  reopening the file
- Move `LHEEvent` to bottom so less quoting needed for typing
- Use `yield from` in `with` blocks to ensure open buffer survives the
  generator lifetime.
- No longer use an intermediate `dict` to improve typing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,6 +23,8 @@ repos:
     rev: v2.4.1
     hooks:
     - id: codespell
+      additional_dependencies:
+        - tomli
 
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.13.3"
--- a/docs/source/examples/02_filter_events_example.ipynb
+++ b/docs/source/examples/02_filter_events_example.ipynb
@@ -1178,14 +1178,6 @@
     "# continue with modified lhe, rivet, pythia ,...\n",
     "plot(\"filtered.lhe.gz\")"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "67370f77-f471-4901-8daf-9f8979fdb4bf",
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,3 +147,8 @@ ignore = [
   "RUF012",   # TODO: mutable class attributes
   "SIM115",   # TODO: use context manager for opening files
 ]
+
+[tool.codespell]
+# https://github.com/codespell-project/codespell/issues/2801
+# Ignore very long base64-like strings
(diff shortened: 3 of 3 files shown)


environment.yml
```diff
@@ -0,0 +1,10 @@
+name: pylhe
+channels:
+  - conda-forge
+dependencies:
+  - hist
```
eduardo-rodrigues: Super minor comment but it would make sense to give at least lower limits to these dependencies as we do for the package itself.

eduardo-rodrigues (approved):
Hi @APN-Pucky, I'm rather loaded but please go ahead and merge when you're happy - I trust you, obviously.

I see there are 2-3 suggestions left.

For the notebook, is that what you will show at PyHEP 2025? Maybe together with some of the other notebooks with a little bit more manipulations?

APN-Pucky:
The new approach of not reopening the files is faster (for reading both init and events, which is the primary use case):

```bash
apn@APNsL13 ~/git/pylhe $ hyperfine "pytest 'tests/test_lhe_reader.py::test_read_lhe_with_attributes_generator' --count 100" --warmup 1
Benchmark 1: pytest 'tests/test_lhe_reader.py::test_read_lhe_with_attributes_generator' --count 100
  Time (mean ± σ):     94.986 s ±  0.501 s    [User: 94.196 s, System: 0.737 s]
  Range (min … max):   94.081 s … 95.712 s    10 runs
 
apn@APNsL13 ~/git/pylhe $ hyperfine "pytest 'tests/test_lhe_reader.py::test_read_lhe_with_attributes_generator' --count 100" --warmup 1
Benchmark 1: pytest 'tests/test_lhe_reader.py::test_read_lhe_with_attributes_generator' --count 100
  Time (mean ± σ):     92.224 s ±  0.762 s    [User: 91.471 s, System: 0.722 s]
  Range (min … max):   91.100 s … 93.376 s    10 runs
```

(Despite blasting deprecation warnings)

PR #332, opened 2025-10-30
chore: Add support for Python 3.14, minor modernisations, more citations of pylhe

eduardo-rodrigues:
Kindly review and merge if all OK and your happy. Thanks @APN-Pucky.


pyproject.toml
```diff
@@ -117,6 +118,18 @@ enable_error_code = [
     "truthy-bool",
 ]
 
+[tool.pytest.ini_options]
+minversion = "6.0"
+testpaths = ["tests"]
+log_cli_level = "INFO"
+xfail_strict = true
+filterwarnings = ["error"]
```
APN-Pucky: I thought `filterwarnings = ["error"]` would make all warnings errors. I remember we had many warnings, so I am a bit surprised that the tests pass.

APN-Pucky: I think our pytest.ini overwrites your new pyproject.toml settings https://github.com/scikit-hep/pylhe/blob/main/pytest.ini

.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.11', '3.12', '3.13', '3.14']
```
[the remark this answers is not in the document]
eduardo-rodrigues: I removed one version to avoid having too many. But agreed that from a strict point a view a given version can only be supported if there are tests for it. So I added 3.10 back.

pyproject.toml
APN-Pucky: Yes, we can enable it, but probably should then filter `.*is deprecated and will be removed in a future version.*`. So something like:

```toml
filterwarnings = [
  "error",
  "ignore:.*is deprecated and will be removed in a future version.*:DeprecationWarning"
]
```

Alternatively, rewriting the tests to not test the old deprecated code would reduce the (backwards-compatibility)  coverage.

eduardo-rodrigues:
This seems now ready to me. I will be away most of the time until the afternoon. Feel free to finalise here and merge. Thanks.

eduardo-rodrigues:
This small PR is finally ready! Let me know @APN-Pucky.

APN-Pucky (approved):

PR #335, opened 2025-10-31
chore: Require graphviz>=0.19 (released late 2021)

eduardo-rodrigues:
I think this bump in min version makes sense 4 years down the line.


eduardo-rodrigues:
And with this trivial update we're at 100% coverage for the first time 👍! (It does not mean we should not make enhancements for "physics oriented tests" rather than code tests.)

PR #370, opened 2026-05-15
fix: Update build command in publish-package.yml

APN-Pucky:
See failure here https://github.com/scikit-hep/pylhe/actions/runs/25913462402/job/76164116377


APN-Pucky:
Publishing 1.0.3 failed because of this. So I try 1.0.4 next after this goes in.

eduardo-rodrigues (approved):
Ah, indeed. Thank you.

PR #372, opened 2026-05-21
feat: Remove all 1.0.0 deprecated functions

APN-Pucky:
- allow `to_awkward` to accept a `LHEFile`.
- Rename `LHEWeightGroup` to `LHEInitRWGTWeightGroup` to avoid confusion with 1.0.0 `LHEWeightGroup` definition
- remove deprecated read_*
- remove deprecated write_*
- deprecate LHEParticle.event and LHEParticle.mothers
- remove Dictcompatibility class


docs/source/examples/00_quickstart.ipynb
```diff
@@ -5,7 +5,7 @@
    "id": "d003df19-502d-42b2-94d6-fa9d909452ff",
    "metadata": {},
    "source": [
-    "# PyLHE version 1.0.0 demo"
+    "# PyLHE version 2.0.0 demo"
```
eduardo-rodrigues, line 8: Apologies @APN-Pucky that I lost track here. So this major update is to then go with a major 2.0 release, in spite of the fact that we are at 1.0.4?

docs/source/examples/01_zpeak.ipynb
```diff
@@ -110,17 +110,17 @@
        "<div style=\"flex=grow:1;\">\n",
        "Regular(30, 50, 150, label='Axis 0')<br/>\n",
        "<hr style=\"margin-top:.2em; margin-bottom:.2em;\"/>\n",
-       "Int64() Σ=26378160.0 <em>(33560000.0 with flow)</em>\n",
+       "Int64() Σ=13189080.0 <em>(16780000.0 with flow)</em>\n",
```
eduardo-rodrigues, line 113: Do you know why these stats related numbers (same below) change this much? The input file is the same, right? Just checking to be sure this is expected.

docs/source/examples/00_quickstart.ipynb
APN-Pucky: Yes, because the way we handle weights is a breaking change so the next release would be 2.0.0 for my side.

docs/source/examples/01_zpeak.ipynb
APN-Pucky: Not sure, but seems to be exactly a factor of 2.

docs/source/examples/00_quickstart.ipynb
eduardo-rodrigues: It makes sense, of course. I vaguely remember some exchanges, but it is always best to check. Thanks.

docs/source/examples/01_zpeak.ipynb
APN-Pucky: I guess that is due to a hist update because the matplotlib representation of the same histogram is still the same:

eduardo-rodrigues: Yep. OK, let's ignore now. The rendering of the plot should visually tells us, together with the file content.

src/pylhe/__init__.py
```diff
@@ -258,7 +167,17 @@ def __post_init__(self) -> None:
 
     @property
     def event(self) -> Optional["LHEEvent"]:
-        """Reference to the parent event, set when the particle is added to an event."""
+        """
+        Reference to the parent event, set when the particle is added to an event.
+
+        .. deprecated:: 2.0.0
+            Access by `particle.event` is deprecated and will be removed in a future version.
```
eduardo-rodrigues, line 174: Suggest the new way as below for mothers?

src/pylhe/__init__.py
```diff
@@ -640,50 +553,32 @@ def _fromcontext(cls, _root: ET.Element, context: Any) -> Union["LHEHeader", Non
...
-        self.attributes["name"] = value
+    extra_attributes: dict[str, str] = field(default_factory=dict)
+    """Generator XML attributes not represented by dedicated fields"""
 
     @property
-    def version(self) -> str:
-        """Version of the generator, retrieved from the attributes."""
-        return self.attributes.get("version", "")
+    def attributes(self) -> dict[str, str]:
+        """Return the attributes of the generator element"""
```
```suggestion
        """Return all the attributes of the generator element, uncluding name and version"""
```

src/pylhe/__init__.py
```diff
@@ -640,50 +553,32 @@ def _fromcontext(cls, _root: ET.Element, context: Any) -> Union["LHEHeader", Non
...
+        return {**self.extra_attributes, "name": self.name, "version": self.version}
```
eduardo-rodrigues, line 579:
```suggestion
        return {"name": self.name, "version": self.version, **self.extra_attributes}
```

src/pylhe/__init__.py
APN-Pucky: There is no new way. Because we don't want to track with every particle anymore to which event it belongs. Because that could cause problems if someone moves particles between events (for whatever reason).

Instead we aim to have a strict hierarchy `Event -> Particle` instead of `Event <-> Particle`, which is easier to handle and makes adding `numba.jit` support easier in the future.

src/pylhe/__init__.py
```diff
@@ -797,7 +650,7 @@ def _fromcontext(cls, _root: ET.Element, context: Any) -> "LHEInit":
                         version=element.attrib["version"],
                         name="",
                         description=element.text.strip() if element.text else "",
-                        attributes=element.attrib.copy(),
+                        extra_attributes=element.attrib.copy(),
```
eduardo-rodrigues, line 656: There is no `def attributes` function as above for example for LHEGenerator? Apologies if I missed it.

src/pylhe/__init__.py
APN-Pucky: I picked this ordering such that if a user wrongly adds "name" or "version" to `extra_attributes`. The correct directly typed `LHEGenerator.name` and `LHEGenerator.version` take precedence.

src/pylhe/__init__.py
APN-Pucky: since LHEGenerator is a `@dataclass` the `attributes` is a read-only (constructed) `@property`. The dataclass constructor only knows about extra_attributes, since the constructor is automatically generated from the fields.

src/pylhe/__init__.py
eduardo-rodrigues: That is a very good point - safety. Thank you for the info.

src/pylhe/__init__.py
eduardo-rodrigues: OK OK. Maybe add such a comment so that nobody can be surprised by the ordering when reading the code.

src/pylhe/__init__.py
eduardo-rodrigues: OK

eduardo-rodrigues (approved):
This looks great and is quite an improvement!

I left some trivial comments/questions, largely my ignorance on the tech side of LHE files. Feel free to resolve everything and merge.

Super, thank you.

APN-Pucky:
Thanks for reviewing it!

Issue #378, opened 2026-06-12
pylhe review from Claude Fable


henryiii:
:robot: _AI text below_ :robot:

The review is done — I read both source modules, the config, and verified the suspicious spots by executing code. Here are the findings, confirmed ones first.

## Bugs (confirmed by execution)

**1. `version` is never populated when reading a file.** In `frombuffer` (src/pylhe/__init__.py:1053), the dataclass is constructed with `version=None`, and `__post_init__` (which extracts `version` from `extra_attributes`) runs at that moment — *before* the generator later assigns `lhef.extra_attributes = root.attrib.copy()` at line 1008. So for a file declaring `version="3.0"`, `lhef.version` is `None` (the value only survives inside `extra_attributes`, which is why `write()` round-trips it and tests don't notice). Fix: set `lhef.version = root.attrib.get("version")` alongside line 1008, or re-run the pop logic there.

**2. `tolhe()` silently drops event attributes and `#` comment lines.** `LHEEvent._fromcontext` carefully parses `<event attr1="foo">` attributes and `#`-prefixed lines into `attributes`/`optional`, but `tolhe()` (line 739) hardcodes `"<event>\n"` and never emits either — verified: round-trip loses both. This contradicts the core round-trip convention in CLAUDE.md. Fix: use `_open_xml_tag("event", self.attributes)` and write the `optional` lines (they conventionally go after the particle lines).

**3. `graph` can draw edges from the wrong mother.** `_build_graph` (line 860) does `self.particles.index(mother)`, which uses dataclass `__eq__` over all 13 fields — when two particles are field-identical (e.g. two identical incoming gluons, which I verified happens), `.index()` returns the first match. Confirmed: a particle with mothers (1, 2) produced `0 -> 2` twice instead of `0 -> 2` and `1 -> 2`. Fix: build edges directly from `mother1`/`mother2` indices instead of identity-searching the list; this also removes an O(n²) scan.

**4. `tofile()` rejects `pathlib.Path`.** `_open_write_file` (line 1133) calls `filepath.endswith(...)` → `AttributeError: 'PosixPath' object has no attribute 'endswith'`, while `fromfile` accepts any `PathLike`. Fix: `filepath = os.fspath(filepath)` first (and widen the annotation to `PathLike`).

[...]

PR #379, opened 2026-06-12
chore: merge pytest.ini into pyproject and adopt PEP 639 license metadata

henryiii:

## Summary

Part of #378

- **Fix pytest config conflict**: `pytest.ini` was silently overriding `[tool.pytest.ini_options]` in `pyproject.toml` (pytest uses the first config file it finds). This meant `minversion`, `testpaths`, `log_level`, `xfail_strict = true`, `--strict-markers`, `--strict-config`, and `-ra` were all inactive. The fix merges the still-relevant options from `pytest.ini` (`junit_family`, coverage addopts, doctest options) into `[tool.pytest.ini_options]` and deletes `pytest.ini`. Stale ignores (`--ignore=setup.py`, `--ignore=binder/`, `--ignore=docs/`) and the pydocstyle-specific `docstyle_convention = numpy` option are dropped. `--cov-config=.coveragerc` is also removed since that file does not exist.

- **PEP 639 license metadata**: Replaces `license = { text = "Apache-2.0" }` with the SPDX expression form `license = "Apache-2.0"` plus `license-files = ["LICENSE"]`. Removes the `"License :: OSI Approved :: Apache Software License"` classifier (forbidden alongside SPDX license expressions per PEP 639). Bumps `hatchling` build requirement to `>=1.27` (first version with full metadata 2.4 / PEP 639 support). The produced wheel now includes `License-Expression: Apache-2.0` in its METADATA.

## Test plan

- [x] `uv run pytest` passes with `configfile: pyproject.toml` confirmed in the session header
- [x] `--strict-config` and `--strict-markers` are now active (no unknown options or unregistered markers)
- [x] 119 tests pass; 2 failures are pre-existing graphviz binary not installed locally (unrelated)
- [x] `uv build` produces a wheel with `License-Expression: Apache-2.0`
- [x] `prek -a --quiet` passes cleanly

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,0 @@
-[pytest]
-docstyle_convention = numpy
-junit_family = xunit2
-addopts = --ignore=setup.py --ignore=binder/ --ignore=docs/ --cov=pylhe --cov-report=term-missing --cov-config=.coveragerc --cov-report xml --doctest-modules --doctest-glob='*.rst'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -129,6 +129,8 @@ addopts = [
     "-ra",
     "--strict-markers",
     "--strict-config",
+    # Non-test dirs that may hold collectable .py/.rst files (e.g. examples/ with future test_* helpers)
+    "--ignore=docs",
     "--cov=pylhe",
     "--cov-report=term-missing",
     "--cov-report=xml",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "hatchling>=1.13.0",
+    "hatchling>=1.27",
     "hatch-vcs>=0.3.0",
 ]
 build-backend = "hatchling.build"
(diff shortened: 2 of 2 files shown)


pytest.ini
```diff
@@ -1,4 +0,0 @@
-[pytest]
-docstyle_convention = numpy
-junit_family = xunit2
-addopts = --ignore=setup.py --ignore=binder/ --ignore=docs/ --cov=pylhe --cov-report=term-missing --cov-config=.coveragerc --cov-report xml --doctest-modules --doctest-glob='*.rst'
```
APN-Pucky, line 4: I think the "stale" `--ignore=docs/` should remains because it also includes the examples folder which might contain test_* .py file functions in the future.

henryiii: :robot: _AI text below_ :robot:

Good catch. Restored the equivalent as `--ignore=docs` in `[tool.pytest.ini_options]` addopts (802b... → latest). Since `testpaths = ["tests"]` already scopes bare `pytest`, the ignore is the belt-and-suspenders guard so `--doctest-glob=*.rst` never recurses into the Sphinx sources under `docs/` (and any future `test_*` helpers elsewhere) when pytest is pointed at the repo root. Left a comment noting the examples-folder rationale.

henryiii: Ahh, didn't ask me about replying. And that's not what I wanted it to do... (Fable is gone, so using Opus now)

henryiii: Ahh, no, I seem to be wrong. norecursedirs is what I was thinking, but it seems like this is supposed to be the one you use for this sort of thing. So it's fine.

eduardo-rodrigues (approved):
Looks good. Feel free to resolve and merge, @APN-Pucky.

APN-Pucky (approved):
Thanks!

PR #380, opened 2026-06-12
fix: correct mother edges in event graph and stop embedding event repr in nodes

henryiii:

## Summary

Fixes three confirmed bugs in `LHEEvent._build_graph` and the `_graph` field. Part of #378.

### Bug 1: wrong edges when two mother particles are field-identical

`_build_graph` used `self.particles.index(mother)` to find a mother's position, but `list.index` uses dataclass `__eq__` across all 13 fields. When two particles are field-identical (e.g. two incoming gluons with the same kinematics), both resolve to index 0, so a daughter particle that has them as distinct mothers gets two edges from index 0 instead of one from index 0 and one from index 1.

**Fix:** use `p.mother1` and `p.mother2` directly as 1-based indices (mirroring the logic already in `LHEEvent.mothers`). This also eliminates the O(n²) scan.

### Bug 2: `attr_dict=str(p.__dict__)` embeds the full event in every node

`p.__dict__` contains `_event`, a back-reference to the entire event (all particles, cached Digraph repr). Passing this as `attr_dict` on every node causes O(n²) DOT-source blowup. `attr_dict` is not a meaningful Graphviz DOT attribute and no tests or downstream code consumed it.

**Fix:** drop the `attr_dict` argument entirely.

### Bug 3: `_graph` participated in `__eq__` and `__repr__`

`_graph: Optional[graphviz.Digraph] = None` was a plain dataclass field. Two semantically identical events compared unequal if one had built its graph, and `repr(event)` leaked `<graphviz.graphs.Digraph object at 0x…>`.

**Fix:** `field(default=None, repr=False, compare=False)`.

## Test plan

- `test_graph_distinct_edges_for_identical_mother_particles`: builds a 3-particle event where particles 0 and 1 are field-identical gluons and particle 2 has mother1=1, mother2=2; asserts both `0 -> 2` and `1 -> 2` appear in the DOT source.
- `test_graph_no_attr_dict_in_source`: asserts `"attr_dict"` does not appear anywhere in the DOT source.
- `test_graph_field_excluded_from_eq_and_repr`: asserts two equal events are still equal after one builds its graph, and that `repr(event)` does not contain `"Digraph"`.
- All existing tests pass (117 pass; 2 pre-existing failures due to missing Graphviz binary in CI).


src/pylhe/__init__.py
```diff
@@ -852,12 +852,13 @@ def _build_graph(self) -> None:
...
-            )
+            self._graph.node(str(i), label=label, texlbl=texlbl)
         for i, p in enumerate(self.particles):
-            for mother in self.mothers(p):
-                self._graph.edge(str(self.particles.index(mother)), str(i))
+            first_idx = int(p.mother1) - 1
+            second_idx = int(p.mother2) - 1
+            for mother_idx in (first_idx, second_idx):
+                if mother_idx >= 0:
+                    self._graph.edge(str(mother_idx), str(i))
```
APN-Pucky: Unnecessary code duplication. instead there should be a mothers_ids function and mothers calls that.

Done — extracted `LHEEvent.mothers_ids(particle)` returning the 1-based-corrected positional indices, and both `_build_graph` and `mothers()` now call it, removing the duplication. (f6c9bcc)

APN-Pucky (approved):

eduardo-rodrigues (approved):

Issue #369, opened 2026-05-14
extension to support LHEH5


APN-Pucky:
LHEH5 is a similar but binary format (See Table 1 https://arxiv.org/pdf/2309.13154 or https://arxiv.org/pdf/1905.05120), but it is still quite new and not very adopted. Let's see if there is interest into it and if it becomes more used.

eduardo-rodrigues:
Have you discussed the enhancement with colleagues as for example  Christian Gutschow? That might be a good way forward to seek advice and check interest. You could even email the HSF Forum about this.

eduardo-rodrigues:
Hello @20DM, allow me to ping you explicitly here in case this sounds of interest to you or others you may know. We would be very happy to receive feedback. Thanks.

APN-Pucky:
The first thing we need is a set of test files from maybe pythia/sherpa or which ever codes already support it.

eduardo-rodrigues:
Sure, something to start with. That would naturally go to https://github.com/scikit-hep/scikit-hep-testdata.

APN-Pucky:
https://www.pythia.org/latest-manual/LHAHDF5.html is a good first reference

APN-Pucky:
Here is one file: https://gitlab.com/Pythia8/releases/-/blob/master/examples/ttbar.hdf5

```txt
~/git/pythia8/examples $ h5ls -r ttbar.hdf5 
/                        Group
/event                   Group
/event/aqcd              Dataset {100}
/event/aqed              Dataset {100}
/event/fscale            Dataset {100}
/event/nparticles        Dataset {100}
/event/pid               Dataset {100}
/event/rscale            Dataset {100}
/event/scale             Dataset {100}
/event/start             Dataset {100}
/event/trials            Dataset {100}
/event/weight            Dataset {100}
/init                    Group
/init/PDFgroupA          Dataset {1}
/init/PDFgroupB          Dataset {1}
/init/PDFsetA            Dataset {1}
/init/PDFsetB            Dataset {1}
/init/beamA              Dataset {1}
/init/beamB              Dataset {1}
/init/energyA            Dataset {1}
/init/energyB            Dataset {1}
/init/numProcesses       Dataset {1}
/init/version            Dataset {3}
/init/weightingStrategy  Dataset {1}
/particle                Group
/particle/color1         Dataset {1200}
/particle/color2         Dataset {1200}
/particle/e              Dataset {1200}
/particle/id             Dataset {1200}
[...]

PR #381, opened 2026-06-12
feat!: accept path-like objects in tofile and replace rwgt/weights booleans with LHEFormat enum

### LINKED ISSUE #385: add LHEFormat instead of bools rwgt and weights
Instead of having two bools rwgt and weights having an enum class 

```py
from enum import Enum

class LHEFormat(enum):
    RWGT=1
    WEIGHTS=2
```

would be the better design.

see-also: https://github.com/scikit-hep/pylhe/pull/381/changes#r3408014654

This might also tie in with HDF5: https://github.com/scikit-hep/pylhe/issues/369.

henryiii:
Closes: https://github.com/scikit-hep/pylhe/issues/385

---

🤖 _AI text below_ 🤖

## Summary

- **Bug 1 (tofile + pathlib.Path):** `_open_write_file` called `filepath.endswith(...)` which raises `AttributeError` for `pathlib.Path` objects. Fixed by converting via `os.fsdecode(os.fspath(filepath))` before the suffix check, and widened the type annotations on `_open_write_file` and `LesHouchesEvents.tofile` from `str` to the existing `PathLike` alias (matching the reading side).

- **API change (weight format selection):** The two booleans `rwgt`/`weights` on `LHEEvent.tolhe`, `LesHouchesEvents.write`, `LesHouchesEvents.tolhe`, and `LesHouchesEvents.tofile` are replaced by a single public `LHEFormat` enum with members `RWGT`, `WEIGHTS`, and `NONE`. This makes the previously-possible contradictory `rwgt=True, weights=True` input unrepresentable, so the both-formats `ValueError` (and its early-validation guard and test) are removed. The default is `LHEFormat.RWGT`, preserving the prior default behaviour. `LHEFormat` is exported in `__all__`. This is a breaking change intended for the unreleased 2.0.0.

  The public methods `LesHouchesEvents.write()`, `LesHouchesEvents.tolhe()`, and `LesHouchesEvents.tofile()` now accept a `format: LHEFormat` keyword argument directly. Internal helpers (`LHEOutputFormat`, `LHEFileFormat`) remain as private implementation details and are no longer part of the public API.

## Test plan

- [x] `test_tofile_accepts_pathlib_path` — writes via `pathlib.Path`, reads back, verifies 1 event
- [x] `test_tofile_accepts_pathlib_path_gz` — writes `.lhe.gz` via `pathlib.Path`, checks gzip magic, reads back via pylhe
- [x] `test_tofile_weights_format` — `tofile(format=LHEFormat.WEIGHTS)` emits a `<weights>` block and no `<rwgt>`
- [x] `test_tofile_none_format_emits_no_weight_block` — `format=LHEFormat.NONE` emits no weight block
- [x] `test_lheformat_is_exported` — `LHEFormat` is in the public API
- [x] Full suite passes with 100% coverage; mypy (strict) clean


src/pylhe/__init__.py
```diff
@@ -920,6 +920,9 @@ def write(
         """
         Write the LHE file to an output stream.
         """
+        if rwgt and weights:
+            err = "Cannot specify both rwgt and weights formats simultaneously."
+            raise ValueError(err)
```
APN-Pucky: Ok, but while we are in the process of releasing 2.0.0 which already breaks some behaviour we could/should switch to having a `LHEFormat` enum instead of two bool that might fail. that way the user can not provide wrong inputs and knows exactly what is possible to get.

Done in 78427b7. Switched to a public `LHEFormat` enum (`RWGT`/`WEIGHTS`/`NONE`) replacing the two `rwgt`/`weights` booleans on `tolhe`/`write`/`tofile`, so the contradictory both-true input is now unrepresentable. The default is `LHEFormat.RWGT`, preserving the prior behaviour, and the now-impossible both-formats `ValueError` (plus its test) has been removed.

henryiii: It decided this made it "feat!". Rename as you see fit. :)

src/pylhe/__init__.py
```diff
@@ -702,27 +712,24 @@ def __post_init__(self) -> None:
...
+            format (LHEFormat): How to serialize event weights (RWGT, WEIGHTS,
+                or NONE).
 
         Returns:
             str: The event as a string in LHE format.
         """
-        if rwgt and weights:
-            err = "Cannot specify both rwgt and weights formats simultaneously."
-            raise ValueError(err)
         sweights = ""
```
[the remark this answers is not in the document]
APN-Pucky: iirc, mypy will not like that

APN-Pucky (approved):

src/pylhe/__init__.py
```diff
@@ -702,27 +712,24 @@ def __post_init__(self) -> None:
...
-    def tolhe(self, rwgt: bool = True, weights: bool = False) -> str:
+    def tolhe(self, format: LHEFormat = LHEFormat.RWGT) -> str:
         """
         Return the event as a string in LHE format.
 
         Args:
-            rwgt (bool): Include the weights in the 'rwgt' format.
-            weights (bool): Include the weights in the 'weights' format.
+            format (LHEFormat): How to serialize event weights (RWGT, WEIGHTS,
+                or NONE).
```
```suggestion
                or NONE), see the `LHEFormat Enum`.
```

src/pylhe/__init__.py
```diff
@@ -915,7 +922,7 @@ def attributes(self) -> dict[str, str]:
         return {**self.extra_attributes, **attrs}
 
     def write(
-        self, output_stream: TWriteable, rwgt: bool = True, weights: bool = False
+        self, output_stream: TWriteable, format: LHEFormat = LHEFormat.RWGT
```
eduardo-rodrigues: Is this the best default format? Just to be sure nobody gets mislead.

src/pylhe/__init__.py
```diff
@@ -927,35 +934,33 @@ def write(
...
     ) -> None:
         """
         Write the LHE file.
 
         Args:
             filepath: Path to the output file.
             gz: Whether to gzip the output file (ignored if filepath suffix is .gz/.gzip).
-            rwgt: Whether to include weights in 'rwgt' format.
-            weights: Whether to include weights in 'weights' format.
+            format: How to serialize event weights (RWGT, WEIGHTS, or NONE).
```
```suggestion
            format: How to serialize event weights (RWGT, WEIGHTS, or NONE), see the `LHEFormat Enum`.
```

tests/test_lhe_writer.py
```diff
@@ -135,7 +137,7 @@ def test_write_lhe():
     init = file.init.tolhe()
 
     assert (
-        file.tolhe(True, False)
+        file.tolhe(pylhe.LHEFormat.RWGT)
```
eduardo-rodrigues:
```suggestion
        file.tolhe(format=pylhe.LHEFormat.RWGT)
```

src/pylhe/__init__.py
APN-Pucky: It is what madgraph uses and it is a bit more expressive than `weights`. Thus I think it is ok to keep it.

eduardo-rodrigues: OK, thanks.

APN-Pucky:
@eduardo-rodrigues we could also design it with a broader scope like:

```py
class LHEFormat:
   version : 1/2/3 
   indent : str = "  "
   weights : rwgt/weights/none
   file: plain/gzip/hdf5
```

which might be more usable for future usages and can be passed through all `tolhe` calls and we could easily extend the code to produce different outputs (i.e. if someone does not want to keep comments/attributes/...)

henryiii:
Ping me if you want me to ask for a rebase and fixes to make CI green.

eduardo-rodrigues:
That would be welcome, @henryiii, as I will have to move to other things today - was only trying to tick off some quick ones but see indeed a CI error, probably from something in the merging that did not go well. Thank you.

As for your comment above, @APN-Pucky, yes, it does sound nice also when thinking about the future as you stated in https://github.com/scikit-hep/pylhe/issues/369.
(Close https://github.com/scikit-hep/pylhe/issues/385 at this point, as done?)

henryiii:
If you need, I'm always happy to trigger Opus to address reviews, I only made it to 23% or something of my weekly quota before the reset. Copilot has been very pricy since June 1! (The code review of Copilot is great, though)

APN-Pucky:
Thanks, I added a extensible `LHEOutputFormat` now with a few predefined defaults for users.

APN-Pucky (approved):

src/pylhe/__init__.py
```diff
@@ -55,6 +65,55 @@ def __dir__() -> list[str]:
 PathLike = Union[str, bytes, os.PathLike[str], os.PathLike[bytes]]
 
 
+class LHEWeightFormat(enum.Enum):
```
eduardo-rodrigues, line 68: All these will need to be well documented at a prominent place in https://scikit-hep.org/pylhe/ so that anyone can immediately know how to use them. Same for any examples and quickstarts. Fair enough that the docs will be a follow-up as this is a very nice and long-ish enhancement.

FYI and eyes @matthewfeickert.

src/pylhe/__init__.py
```diff
@@ -55,6 +65,55 @@ def __dir__() -> list[str]:
...
+"""Output format with RWGT weight block."""
+WEIGHTS_FORMAT = LHEOutputFormat(weights=LHEWeightFormat.WEIGHTS)
+"""Output format with WEIGHTS weight block."""
+RWGT_GZ_FORMAT = LHEOutputFormat(weights=LHEWeightFormat.RWGT, file=LHEFileFormat.GZIP)
+"""Output format with RWGT weight block and gzip compressed file format."""
+WEIGHTS_GZ_FORMAT = LHEOutputFormat(
+    weights=LHEWeightFormat.WEIGHTS, file=LHEFileFormat.GZIP
+)
+"""Output format with WEIGHTS weight block and gzip compressed file format."""
+NO_WEIGHTS_FORMAT = LHEOutputFormat(weights=LHEWeightFormat.NONE)
```
eduardo-rodrigues, line 114:
```suggestion
NO_WEIGHTS_FORMAT = LHEOutputFormat(weights=LHEWeightFormat.NONE)
"""Output format with no WEIGHTS weight block and (default) plain text file format.""
```

src/pylhe/__init__.py
```diff
@@ -55,6 +65,55 @@ def __dir__() -> list[str]:
...
+
+# User convenience presets for common formats
+DEFAULT_FORMAT = LHEOutputFormat()
+"""Default output format with indentation, RWGT weight block and plain text file format."""
+GZIP_FORMAT = LHEOutputFormat(file=LHEFileFormat.GZIP)
+"""Output format for gzip compressed files."""
+RWGT_FORMAT = LHEOutputFormat(weights=LHEWeightFormat.RWGT)
+"""Output format with RWGT weight block."""
+WEIGHTS_FORMAT = LHEOutputFormat(weights=LHEWeightFormat.WEIGHTS)
+"""Output format with WEIGHTS weight block."""
```
```suggestion
"""Output format with WEIGHTS weight block and (default) plain text file format."""
```

src/pylhe/__init__.py
```diff
@@ -55,6 +65,55 @@ def __dir__() -> list[str]:
...
+    procinfo: str = "{xSection: 14.7e} {error: 14.7e} {unitWeight: 14.7e} {procId: 5d}"
+
+
+# User convenience presets for common formats
+DEFAULT_FORMAT = LHEOutputFormat()
+"""Default output format with indentation, RWGT weight block and plain text file format."""
+GZIP_FORMAT = LHEOutputFormat(file=LHEFileFormat.GZIP)
+"""Output format for gzip compressed files."""
+RWGT_FORMAT = LHEOutputFormat(weights=LHEWeightFormat.RWGT)
+"""Output format with RWGT weight block."""
```
```suggestion
"""Output format with RWGT weight block and (default) plain text file format."""
```

src/pylhe/__init__.py
```diff
@@ -55,6 +65,55 @@ def __dir__() -> list[str]:
...
+    particle: str = "{id:5d} {status:3d} {mother1:3d} {mother2:3d} {color1:3d} {color2:3d} {px: 15.8e} {py: 15.8e} {pz: 15.8e} {e: 15.8e} {m: 15.8e} {lifetime: 10.4e} {spin: 10.4e}"
+    initinfo: str = " {beamA: 6d} {beamB: 6d} {energyA: 14.7e} {energyB: 14.7e} {PDFgroupA: 5d} {PDFgroupB: 5d} {PDFsetA: 5d} {PDFsetB: 5d} {weightingStrategy: 5d} {numProcesses: 5d}"
+    procinfo: str = "{xSection: 14.7e} {error: 14.7e} {unitWeight: 14.7e} {procId: 5d}"
+
+
+# User convenience presets for common formats
+DEFAULT_FORMAT = LHEOutputFormat()
+"""Default output format with indentation, RWGT weight block and plain text file format."""
+GZIP_FORMAT = LHEOutputFormat(file=LHEFileFormat.GZIP)
+"""Output format for gzip compressed files."""
```
```suggestion
"""Output format for gzip compressed files, with (default) RWGT weight block."""
```

src/pylhe/__init__.py
```diff
@@ -55,6 +65,55 @@ def __dir__() -> list[str]:
...
+
+    eventinfo: str = "{nparticles:3d} {pid:6d} {weight: 15.10e} {scale: 15.10e} {aqed: 15.10e} {aqcd: 15.10e}"
+    particle: str = "{id:5d} {status:3d} {mother1:3d} {mother2:3d} {color1:3d} {color2:3d} {px: 15.8e} {py: 15.8e} {pz: 15.8e} {e: 15.8e} {m: 15.8e} {lifetime: 10.4e} {spin: 10.4e}"
+    initinfo: str = " {beamA: 6d} {beamB: 6d} {energyA: 14.7e} {energyB: 14.7e} {PDFgroupA: 5d} {PDFgroupB: 5d} {PDFsetA: 5d} {PDFsetB: 5d} {weightingStrategy: 5d} {numProcesses: 5d}"
+    procinfo: str = "{xSection: 14.7e} {error: 14.7e} {unitWeight: 14.7e} {procId: 5d}"
+
+
+# User convenience presets for common formats
+DEFAULT_FORMAT = LHEOutputFormat()
+"""Default output format with indentation, RWGT weight block and plain text file format."""
```
eduardo-rodrigues: I did not change everywhere but it probably makes more sense to say "weightS block" everywhere.

src/pylhe/__init__.py
```diff
@@ -55,6 +65,55 @@ def __dir__() -> list[str]:
...
+    WEIGHTS = "weights"  # <weights>...</weights> block
+    NONE = "none"  # no weight block emitted
+
+
+class LHEFileFormat(enum.Enum):
+    """Selects the file format."""
+
+    PLAIN = "plain"
+    GZIP = "gzip"
+    # HDF5 = "hdf5" # TODO
```
```suggestion
    # HDF5 = "hdf5" # TODO - for future support of LHEH5, see https://github.com/scikit-hep/pylhe/issues/369
```

src/pylhe/__init__.py
```diff
@@ -681,27 +777,23 @@ def __post_init__(self) -> None:
...
 
-    def tolhe(self, rwgt: bool = True, weights: bool = False) -> str:
+    def tolhe(self, lheformat: LHEOutputFormat = DEFAULT_FORMAT) -> str:
         """
         Return the event as a string in LHE format.
 
         Args:
-            rwgt (bool): Include the weights in the 'rwgt' format.
-            weights (bool): Include the weights in the 'weights' format.
+            format (LHEOutputFormat): How to serialize event.
```
```suggestion
            format (LHEOutputFormat): How to serialize the event, see the `LHEOutputFormat Enum`.
```

src/pylhe/__init__.py
```diff
@@ -928,38 +1020,35 @@ def write(
...
     ) -> None:
         """
         Write the LHE file.
 
         Args:
             filepath: Path to the output file.
             gz: Whether to gzip the output file (ignored if filepath suffix is .gz/.gzip).
-            rwgt: Whether to include weights in 'rwgt' format.
-            weights: Whether to include weights in 'weights' format.
+            format: How to serialize event weights (RWGT, WEIGHTS, or NONE), see the `LHEOutputFormat Enum`.
```
```suggestion
            format (LHEOutputFormat): How to serialize the event, see the `LHEOutputFormat Enum`.
```

APN-Pucky: I think this should also be `lheformat` as variable name (to avoid conflicts with builtin `format`)

src/pylhe/__init__.py
```diff
@@ -928,38 +1020,35 @@ def write(
...
-        rwgt: bool = True,
-        weights: bool = False,
+        filepath: PathLike,
+        lheformat: LHEOutputFormat = DEFAULT_FORMAT,
     ) -> None:
         """
         Write the LHE file.
 
         Args:
             filepath: Path to the output file.
```
```suggestion
            filepath (PathLike): Path to the output file.
```

src/pylhe/__init__.py
```diff
@@ -928,38 +1020,35 @@ def write(
...
             gz: Whether to gzip the output file (ignored if filepath suffix is .gz/.gzip).
```
```suggestion
```

eduardo-rodrigues (approved):
Hi @APN-Pucky, I tried my best to review this yet another very nice enhancement. Congrats!

I left some easy suggestions. Already approving.

Thank you very much.

src/pylhe/__init__.py
eduardo-rodrigues: Agreed, sorry I did not spot the change of name (convention).

src/pylhe/__init__.py
eduardo-rodrigues: I see you followed the suggestion. Resolving.

APN-Pucky:
Thanks @eduardo-rodrigues and @henryiii

PR #382, opened 2026-06-12
perf: import awkward/vector/graphviz/particle lazily

henryiii:

## Summary

`import pylhe` previously eagerly imported four heavy dependencies that the core reader/writer never needs:

- **awkward** and **vector** — only needed for `to_awkward` (pulled in by the module-bottom `from .awkward import to_awkward`, which also runs `vector.register_awkward()` and several `ak.mixin_class` registrations at import time).
- **graphviz** and **particle** — only needed for the `LHEEvent.graph` visualization (plus the module-level `_PDGID2LaTeXNameMap = DirectionalMaps(...)` construction).

This PR makes all four lazy so the core parse path imports nearly instantly, with **no public API changes**.

## Changes

- Removed the bottom-of-module `from .awkward import to_awkward`; `to_awkward` is now exposed via a module-level `__getattr__` that imports it from `.awkward` on first access (raising the standard `AttributeError` for unknown names). `"to_awkward"` stays in `__all__`/`__dir__`. `from pylhe import to_awkward`, `pylhe.to_awkward`, and `import pylhe.awkward` all still work, and the vector/awkward behavior registrations now happen only when the awkward bridge is actually used.
- Moved the `graphviz`/`particle` imports and the `DirectionalMaps` construction into `LHEEvent._build_graph`. The PDGID→LaTeX map is built once and cached via `functools.cache`. Type annotations are quoted with a `TYPE_CHECKING`-guarded `import graphviz`.
- mypy stays green under `strict` (the `# type: ignore[import-untyped]` for untyped graphviz now lives on the `TYPE_CHECKING` import; no stale ignores).

## Laziness verification

```
python -c "import sys, pylhe; assert not any(m in sys.modules for m in (\"awkward\",\"vector\",\"graphviz\",\"particle\"))"
```

passes after the change. A new `tests/test_api.py::test_lazy_heavy_imports` asserts this in a **fresh interpreter** (via `subprocess.run`) so other tests' imports do not pollute `sys.modules`. `pylhe.to_awkward` and `LHEEvent.graph` were verified end-to-end (covered by `tests/test_awkward.py` and `tests/test_visualize.py`).

## Notes

- Dependencies are intentionally **not** changed: graphviz/particle/awkward/vector remain required deps. Making them optional extras is a possible follow-up, deliberately out of scope here.

Part of #378

### DIFF
--- a/src/pylhe/__init__.py
+++ b/src/pylhe/__init__.py
@@ -2,6 +2,7 @@
 Python interface to read Les Houches Event (LHE) files.
 """
 
+import functools
 import gzip
 import io
 import os
@@ -11,6 +12,7 @@ from collections.abc import Iterable, Iterator
 from copy import deepcopy
 from dataclasses import dataclass, field
 from typing import (
+    TYPE_CHECKING,
     Any,
     BinaryIO,
     Optional,
@@ -21,12 +23,10 @@ from typing import (
 )
 from xml.sax.saxutils import quoteattr
 
-import graphviz  # type: ignore[import-untyped]
-from particle import latex_to_html_name
-from particle.converters.bimap import DirectionalMaps
-from particle.exceptions import MatchingIDNotFound
-
 from pylhe._version import version as __version__
+
+if TYPE_CHECKING:
+    import graphviz  # type: ignore[import-untyped]
 
 __all__ = [
     "LHEEvent",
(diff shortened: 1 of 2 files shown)


APN-Pucky:
I don't like any of these changes. The code becomes more ugly, especially importing graphviz twice conditionally ...

The import time is negligible IMO. 

Also test tests are failing :-1:

henryiii:
Don't do this, you should add `__lazy_modules__` (my flake8-lazy tool will do that for you if you want). You'll only get savings on 3.15+ (so no rush, binaries won't come until after the RC).

henryiii:
(I can have it babysit tests and make them pass. But don't manually make imports lazy for 3.14 and older, I missed that was one of the suggestions.)

PR #383, opened 2026-06-12
fix: preserve version, event attributes, and comments on round-trip

henryiii:

Fixes three confirmed round-trip / parsing fidelity bugs in `src/pylhe/__init__.py`.

### Bug 1 — `version` field never populated when reading
`LesHouchesEvents.__post_init__` pops `"version"` out of `extra_attributes` into the `version` field, but that runs at construction time, *before* the streaming generator assigns `lhef.extra_attributes = root.attrib.copy()`. So for `<LesHouchesEvents version="3.0">`, `lhef.version` stayed `None` (the value only survived inside `extra_attributes`).

**Fix:** where `extra_attributes` is assigned in the generator, mirror `__post_init__` — pop `"version"` out of the copied dict into `lhef.version` (only when `version` is still `None`). `attributes` still emits `version` exactly once.

### Bug 2 — `LHEEvent.tolhe()` dropped event attributes and `#` comment lines
`_fromcontext` parses `<event attr="...">` attributes into `self.attributes` and `#`-prefixed lines into `self.optional`, but `tolhe()` hardcoded `"<event>\n"` and never emitted either — silent data loss on round-trip.

**Fix:** use `_open_xml_tag("event", self.attributes)` for the opening tag and emit the `optional` lines after the particle lines, before the weights/scales blocks (matching real generator output).

### Bug 3 — live (un-copied) `element.attrib` references
Three sites stored a reference to a live ElementTree `attrib` dict that is later `element.clear()`-ed or mutated by a child `__post_init__` `pop`. Under CPython's C accelerator `clear()` swaps the dict so it happens to work, but under pure-Python ElementTree it clears in place and corrupts already-yielded objects.

**Fix:** use `.copy()` at each site — `LHEEvent._fromcontext` (`attrib`), and the `LHEInitRWGTWeight` / `LHEInitRWGTWeightGroup` construction in `LHEHeader._fromcontext`.

### Tests
- New round-trip tests in `tests/test_lhe_reader.py`: `version` populated after `fromstring` and `fromfile`; event attributes and `#` comments survive `parse -> tolhe -> parse`.
- Updated the expected-output strings in `tests/test_lhe_writer.py`: they now include the v3 test file's `npLO`/`npNLO` event attributes and `#aMCatNLO` comment lines that were previously dropped (i.e. the fix working).

`uv run pytest` (127 passed, incl. doctests), `uv run mypy` (clean, 100% src coverage), and `prek -a` all pass.

Part of #378


src/pylhe/__init__.py
```diff
@@ -736,12 +737,18 @@ def tolhe(self, rwgt: bool = True, weights: bool = False) -> str:
                 + "/>\n"
             )
 
+        soptional = ""
+        if self.optional:
+            soptional = "\n".join(self.optional) + "\n"
+
         return (
-            "<event>\n"
+            _open_xml_tag("event", self.attributes)
```
APN-Pucky, line 745: good

src/pylhe/__init__.py
```diff
@@ -1006,6 +1013,11 @@ def _generator(lhef: LHEFile) -> Iterator[LHEEvent]:
                         raise ValueError(err)
                     else:
                         lhef.extra_attributes = root.attrib.copy()
+                        # __post_init__ ran at construction with no attributes,
+                        # so mirror its logic here to populate the version field.
+                        version = lhef.extra_attributes.pop("version", None)
+                        if lhef.version is None:
+                            lhef.version = version
```
APN-Pucky: Instead of duplicating that code we should just trigger a __post_init__, since there might be more attributes in the future.

Replaced the duplicated version-extraction block with a single `lhef.__post_init__()` call right after `lhef.extra_attributes = root.attrib.copy()`. This re-runs the dataclass post-init against the now-populated attributes dict, so `version` (and any future extra-attribute handling added to `__post_init__`) is populated without duplication — the existing `if self.version is None` guard keeps it idempotent. Done in commit 693a401.

APN-Pucky (approved):

eduardo-rodrigues (approved):
Hi @henryiii and @APN-Pucky, I see that you have reviewed and approved this. Looks good. Merging, thank you very much.

PR #384, opened 2026-06-12
refactor: simplify parsing internals and fix duplicate deprecation warnings

henryiii:

## Summary

Internal-only cleanups to `src/pylhe/__init__.py` — no public API changes.

- **Cleanup 1 — collapse duplicate generator-parsing branches**: `LHEInit._fromcontext` had two near-identical branches for lhe-v3 (name + version) and lhe-v2 (version only). Collapsed into a single `LHEGenerator(name=element.attrib.get("name", ""), ...)` construction; behavior is identical.
- **Cleanup 2 — replace `_indent` with stdlib `ET.indent`**: Deleted the Stack Overflow copy-paste `_indent` helper (added pre-3.9) and replaced both call sites with `ET.indent(root, space="  ")` + `root.tail = "\n"`. Output is byte-identical; confirmed by existing writer tests.
- **Cleanup 3 — annotation and dead-logic fixes**:
  - `LHEHeader._fromcontext` return type narrowed from `Union["LHEHeader", None]` to `"LHEHeader"` (it never returns None).
  - All three `_fromcontext` `context` parameters typed from `Any` to `Iterator[tuple[str, ET.Element]]`.
  - Removed dead `if w` guard in `<weights>` parsing (items from `str.split()` are never falsy).
  - Replaced bare `KeyError` when `<weights>` has more entries than `<initrwgt>` declares with a clear `ValueError` including entry counts.
- **Cleanup 4 — `mothers()` emits exactly one warning**: `LHEParticle.mothers()` was internally accessing `self.event` (the deprecated property) three times, emitting 3 extra `DeprecationWarning`s per call. Changed to use `self._event` directly. Test updated to assert exactly one warning per call.

Part of #378

## Test plan

- [x] `uv run pytest` — 121 passed, 0 failed
- [x] `uv run mypy` — no issues
- [x] `prek -a --quiet` — all hooks pass


src/pylhe/__init__.py
```diff
@@ -232,35 +232,16 @@ def mothers(self) -> list["LHEParticle"]:
...
-            elem.tail = i
-        for inner_elem in elem:
-            _indent(inner_elem, level + 1)
-        elem = inner_elem
-        if not elem.tail or not elem.tail.strip():
-            elem.tail = i
-    elif level and (not elem.tail or not elem.tail.strip()):
-        elem.tail = i
-
-
```
APN-Pucky, line 263: Ok, since we do not support python <3.9 anymore.

src/pylhe/__init__.py
```diff
@@ -458,7 +439,8 @@ def tolhe(self) -> str:
             else:
                 weight_elem = ET.SubElement(root, "weight", attrib=e.attributes)
                 weight_elem.text = e.name
-        _indent(root)
+        ET.indent(root, space="  ")
+        root.tail = "\n"
```
APN-Pucky: I think we should move this to the `_indent` function and keep it. Then, if we decide to change the indenting, or expose it as an option it only needs to be changed in one place.

APN-Pucky (changes requested):

APN-Pucky (approved):

eduardo-rodrigues (approved):
Looks good to me too. Thank you Henry.

PR #394, opened 2026-07-10
feat: use slots=True in dataclasses and remove bidirectional event reference from LHEParticle

### LINKED ISSUE #368: Removal of bidirectional relationship between event and particles
I think we should remove the reference `.event` in each of the particles. 

https://github.com/scikit-hep/pylhe/blob/b00236c684ad9b443495190a1d70ea5bd555b4e5/src/pylhe/__init__.py#L596

https://github.com/scikit-hep/pylhe/blob/b00236c684ad9b443495190a1d70ea5bd555b4e5/src/pylhe/__init__.py#L235-L250

If a user moves particles around between events that information will be wrong, unless the user adjusts it manually. Further, this seems to account for about >1% of the performance profiling I did here https://github.com/scikit-hep/pylhe/pull/367. Removing it should definitely make things faster and makes the library closer to the LHE format.

I do not think anybody uses this variable, but maybe pylhe already passes https://www.hyrumslaw.com/.

### LINKED ISSUE #386: consider adding slots=True to dataclasses
This prevents users from adding/typoing new non existing attributes of the classes. Also might give a speedup.

APN-Pucky:
~4-6% faster reading and ~3-4.5% faster writing

Closes: https://github.com/scikit-hep/pylhe/issues/386
Closes: https://github.com/scikit-hep/pylhe/issues/368

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Only determined .lhe.gz compression from file extension when no LHEOutputFormat is provided. Otherwise, the compression is determined by the LHEOutputFormat.
 - Weights and weightgroups are now no longer stored in `LHEInit.weightgroup` but in `LHEHeader.initrwgt.entries`. They are thus part of the `<header>` instead of `<init>` as demanded by LHE specification.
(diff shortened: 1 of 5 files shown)


eduardo-rodrigues (approved):
The title of the PR needs to be updated as the other update is even more important ;-). Otherwise, this is good.

Best wishes.

PR #396, opened 2026-07-13
chore: Final preparations for release 2.0.0

APN-Pucky:
I am not sure if we should also update the paper to version 2.0.0 since it is now close to being reviewed? Should we already release version 2.0.0? That might confuse the reviewer(s).

Changes here:

- Always set LHE version to 3.0 when writing a file (rivet/hepmc crashes on LHE file without version tag) 
- update changelog for version 2.0.0
- cleaned up old version 1.0.0 demo a bit

I would make the 2.0.0 release right after we merge this.


src/pylhe/__init__.py
```diff
@@ -975,7 +975,9 @@ class LesHouchesEvents:
     """Header block"""
     comment: str | None = None
     """Comment block"""
-    version: str | None = None
+    version: str | None = (
```
eduardo-rodrigues: Why the parentheses here?

src/pylhe/__init__.py
```diff
@@ -1009,7 +1011,15 @@ def write(
             TWriteable: The output stream with the LHE file written to it.
 
         """
-        output_stream.write(_open_xml_tag("LesHouchesEvents", self.attributes) + "\n")
+        write_attributes = self.attributes.copy()
+        if write_attributes.get("version") != "3.0":
+            warnings.warn(
+                f"Writing LHE file with version {write_attributes.get('version')} but pylhe always writes LHE v3.0 files.",
```
eduardo-rodrigues: Then I would add this info to the write method docstring as in "Write the LHE file to an output stream." and wherever else relevant.

src/pylhe/__init__.py
APN-Pucky: Yes, weird... The autoformatter added it because of the comment.

src/pylhe/__init__.py
```diff
@@ -1115,6 +1125,7 @@ def _hdf5_generator() -> Iterator[LHEEvent]:
             return LesHouchesEvents(
                 init=init,
                 events=events if generator else list(events),
+                version="3.0",  # We set version 3.0 here since LHEH5 V2 has scales
```
eduardo-rodrigues: I'm confused here by the setting of version 3 and the comment. In the lheh5.py module the columns are for version 2 only, hence it seems we are only supporting v2. But here you set to 3 and say "since" v2 has scales, which is not a requirement of the method arguments/signature. I'm lost, even if no expert on the format at all. Sorry.

src/pylhe/__init__.py
eduardo-rodrigues: If the formatter is stubbord, one can move the comment to the line above.

eduardo-rodrigues:
> I am not sure if we should also update the paper to version 2.0.0 since it is now close to being reviewed? Should we already release version 2.0.0? That might confuse the reviewer(s).

I think @matthewfeickert will know best here. I don't think we will be done in a few weeks at this stage, so having the paper updated may be the best and less confusion way forward provided we inform the editor about the major v2 release.

> Changes here:
> 
> * Always set LHE version to 3.0 when writing a file (rivet/hepmc crashes on LHE file without version tag)
> * update changelog for version 2.0.0
> * cleaned up old version 1.0.0 demo a bit

This reminds me that the nice table in the README should be updated to reflect the v2 functionality - sorry for this extra burden.

> I would make the 2.0.0 release right after we merge this.

Sounds great, modulo the comment above 👍.

src/pylhe/__init__.py
APN-Pucky: In chronological appearance order there are 5 version (modulo minor pepper changes):

LHE XML:
- 1.0
- 2.0
- 3.0

 LHEH5:
 - 1.0
 - 2.0
 
 but the LHEH5 versioning is unrelated to LHE XML version.

src/pylhe/__init__.py
APN-Pucky: I removed the warning and instead added a proper versioning `LHEVersion.V3 `enum in the `LHEXMLFormat`.
Adding also `LHEVersion.V1` was easy, because it is just excluding scales, generator, weights, etc.

src/pylhe/__init__.py
APN-Pucky: I will revert this change because it is confusing which version is set in `LesHouchesEvents` and with above changes of adding `LHEVersion` to the Formatting, defaulting to version 3. it is no longer a problem if we leave it undefined here.

src/pylhe/__init__.py
eduardo-rodrigues: Yeah, I think that's much better and clearer, thank you.

matthewfeickert:
> > I am not sure if we should also update the paper to version 2.0.0 since it is now close to being reviewed? Should we already release version 2.0.0? That might confuse the reviewer(s).
> 
> I think @matthewfeickert will know best here. I don't think we will be done in a few weeks at this stage, so having the paper updated may be the best and less confusion way forward provided we inform the editor about the major v2 release.

We should update the paper. The reviewers will be reviewing _the repository_ and we will be asked to make updates to the release metadata for the paper anyway. The paper is much more of a living document for the review, so it is fine to move forward with a release update. (Sorry for the slow reply again.)

eduardo-rodrigues:
Hey @APN-Pucky, I think these last updates are clear improvements. I appreciate that versioning is tricky, especially with 2 formats being handled now. In the future, post the v2 release, it will be a good idea to deprecate old stuff as for example the good-old XML format V1.

(Trivial comment but the PR header info needs an update from the review.)

I will do a final pass tomorrow as it's tight now. Unless @matthewfeickert can help finalise.

APN-Pucky:
Ok, thanks! I will update the paper today in this PR.

docs/source/lhe.rst
```diff
@@ -108,5 +182,4 @@ The table below summarizes the main parameters found in LHE files grouped by the
 
 Further details can be found in the original definition of the LHE file standard.
 Besides the original publication there were two extensions to the LHE format, version 2.0 in 2009 :cite:`Butterworth:2010ym` and version 3.0 in 2013 :cite:`Andersen:2014efa`.
-However, ``pylhe`` currently only implements the widely adopted extension from version 1.0, that is the addition of multiple weights via ``<initrwgt>``, ``<rwgt>``, ``<weight>``, ``<weights>``, ``<wgt>``, and ``<weightgroup>``.
-If in the future there is a demand for ``<scales>``, ``<generator>``, ``<pdfinfo>``, or ``<clustering>``, support for these can be added as well.
+However, ``pylhe`` implements the widely adopted version 3.0 and thus also the subset version 1.0, that is the addition ``<scales>``, ``<generator>``, and of multiple weights via ``<initrwgt>``, ``<rwgt>``, ``<weight>``, ``<weights>``, ``<wgt>``, and ``<weightgroup>``.
```
eduardo-rodrigues:
```suggestion
However, ``pylhe`` implements the widely adopted version 3.0 and thus also the subset version 1.0, that is the addition of ``<scales>``, ``<generator>``, and of multiple weights via ``<initrwgt>``, ``<rwgt>``, ``<weight>``, ``<weights>``, ``<wgt>``, and ``<weightgroup>``.
```

paper/paper.md
```diff
@@ -46,13 +46,13 @@ Typically, LHE files are used to describe parton-level events generated by matri
 In contrast to LHE files, HepMC stores fewer details about the particles, but also contains many more particles per event.
 Both files are plain text-based formats, making them human-readable but potentially large.
 Consequently, compression using gzip is common practice for both formats.
-In the future a binary format such as HDF5 [@hdf5] could be considered for event storage to improve read/write performance and reduce file sizes further.
+Since `pylhe` version 2.0.0, the library also supports the HDF5-based LHEH5 format [@hdf5;@Hoche:2019flt;@Bothmann:2023ozs], improving read/write performance and further reducing file sizes for event storage.
```
eduardo-rodrigues:
```suggestion
Since `pylhe` version 2.0.0, the library also supports the HDF5-based [@hdf5] LHEH5 format [@Hoche:2019flt;@Bothmann:2023ozs], improving read/write performance and further reducing file sizes for event storage.
```

paper/paper.md
```diff
@@ -46,13 +46,13 @@ Typically, LHE files are used to describe parton-level events generated by matri
...
 Further details can be found in the original definition of the Les Houches Event file standard [@Alwall:2006yp].
 Following the original publication there were two extensions to the LHE format, version 2.0 in 2009 [@Butterworth:2010ym] and version 3.0 in 2013 [@Andersen:2014efa].
-However, `pylhe` currently only implements the widely adopted extension on top of version 1.0, that is the addition of multiple weights via `<initrwgt>`, `<rwgt>`, `<weight>`, `<weights>`, `<wgt>`, and `<weightgroup>`.
-In the future, if there is a demand for `<scales>`, `<generator>`, `<pdfinfo>`, or `<clustering>`, support for these can be added as well.
+`pylhe` completely implements the widely adopted version 3.0, that is the addition of `<scales>`, `<generator>`, and multiple weights via `<initrwgt>`, `<rwgt>`, `<weight>`, `<weights>`, `<wgt>`, and `<weightgroup>` on top of the version 1.0.
+Features that were introduced in version 2.0 but removed in version 3.0 again, such as `<pdfinfo>` or `<clustering>` are not supported.
```
eduardo-rodrigues:
```suggestion
Features that were introduced in version 2.0 but removed in version 3.0, such as `<pdfinfo>` or `<clustering>`, are not supported.
```

paper/paper.md
```diff
@@ -66,13 +66,14 @@ Unlike the existing C/C++/Fortran interfaces provided by the Monte Carlo event g
 When `pylhe` was first developed, there were no other Python libraries available for reading and writing LHE files.
 Nowadays, there are a few other smaller Python libraries with less adoption than `pylhe`, which provide only read functionality and are no longer actively maintained, such as `lhereader` [@lhereader1;@lhereader2].
 For completeness, it should be mentioned that several LHE libraries exist in other programming languages, such as Go (`go-hep` [@Binet2017]), Rust (`lhe` [@lhe-rs], `lhef` [@lhef;@lhef-rs], `event_file_reader` [@event_file_reader]), Julia (`LHEF.jl` [@LHEF_jl]) and Haskell (`lhe.hs` [@lhehs]).
-These provide varying degrees of completeness.
+These provide varying degrees of completeness with none supporting LHEH5 yet.
```
eduardo-rodrigues:
```suggestion
These provide varying degrees of completeness, but none support the LHEH5 format yet.
```

paper/paper.md
```diff
@@ -66,13 +66,14 @@ Unlike the existing C/C++/Fortran interfaces provided by the Monte Carlo event g
...
 
 # Software design
 
 `pylhe` allows for easy reading and writing of `.lhe` and `.lhe.gz` files in Python, enabling seamless integration into modern data analysis workflows in high-energy physics.
 The pythonic event yielding approach allows for memory-efficient processing of arbitrarily large LHE files by streaming events one at a time rather than loading all of them at once into memory.
 Internally, `pylhe` uses `xml.etree.ElementTree` to parse the XML structure, since using the `lxml` library did not provide a significant speed up.
+For LHEH5 files the library uses `h5py` [@collette2013pythonhdf5] to read and write the HDF5 format, via the same pythonic data structures making it easy to transparently convert between XML and HDF5 formats.
```
eduardo-rodrigues:
```suggestion
For LHEH5 files the library uses `h5py` [@collette2013pythonhdf5] to read and write the HDF5 format, via the same pythonic data structures, making it easy to convert between the standard XML and newer HDF5 formats.
```

src/pylhe/__init__.py
```diff
@@ -78,11 +78,19 @@ class LHEWeightFormat(enum.Enum):
     NONE = "none"  # no weights block emitted
 
 
+class LHEVersion(enum.Enum):
+    """Selects the LHE XML version."""
```
```suggestion
    """
    Selects the LHE XML version.
    
    Note:
        Version 2 is not supported as of now since most attributes have been removed in Version 3, which is the present default.
    """
```

src/pylhe/__init__.py
```diff
@@ -975,7 +1000,7 @@ class LesHouchesEvents:
     """Header block"""
     comment: str | None = None
     """Comment block"""
-    version: str | None = None
+    version: str | None = None  # None for undefined or LHEH5
```
eduardo-rodrigues: I would move the line comment to the attribute docstring just below, so that the documentation renders it to users.

src/pylhe/__init__.py
```diff
@@ -1036,7 +1065,7 @@ def tofile(
         | None = None,  # default format is None because we do file name suffix detection in _open_write_file
     ) -> None:
         """
-        Write the LHE file.
+        Write the LHE file as LHE.
```
eduardo-rodrigues, line 1073:
```suggestion
        Write the LHE file as LHE XML.
```

APN-Pucky: depending on filepath/format we can write either LHE XML or LHE HDF5 here.

eduardo-rodrigues (approved):
Hey @APN-Pucky, I left a few final suggestions. I hope I suggested nothing silly as I'm pretty loaded with work today, and doing as quickly as possible.

In any case, it looks great. Thank you!

src/pylhe/__init__.py
eduardo-rodrigues: Ah, overlooked that, see my comment below :D. Resolving.

APN-Pucky:
Thank you again @eduardo-rodrigues ! After the version 2.0.0 release I think it is pretty complete and only fixes/chore should follow for quiet some time.

